PR #32, opened 2025-03-23
Update README to document class name labels in prediction output

farukalamai:
# Description

This PR updates the README.md to document that predictions now display class names instead of class IDs in bounding box labels. This enhancement improves the user experience by making detection results more human-readable and interpretable.
The documentation now clearly explains that when users run predictions with the model, object classes will be shown by their descriptive names (e.g., "person", "car") rather than numeric IDs, making it easier to understand detection results at a glance.
This change complements the implementation that converts class IDs to class names in the detection output, creating a more intuitive experience for users of the library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,7 @@ import requests
 import supervision as sv
 from PIL import Image
 from rfdetr import RFDETRBase
+from rfdetr.util.coco_classes import COCO_CLASSES
 
 model = RFDETRBase()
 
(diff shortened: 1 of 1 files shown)


SkalskiP:
I initially wanted to keep it extra short, but given how many people asked about that over the past few days I think we should add it in. 👍🏻

SkalskiP (approved):

PR #58, opened 2025-03-26
save metrics plot once the training is completed

SkalskiP:
# Description

Added a new class `MetricsPlotSink` that uses the existing callback system to store and save metric values. After training is complete, the loss values obtained at the end of each epoch for the train set and validation set are plotted, as well as the mAP and mAR values for the base and EMA models.

To unlock this feature, a new callback type `on_train_end` has been added.

example plot

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,6 +60,7 @@ dependencies = [
     "rf100vl",
     "pydantic",
     "supervision",
+    "matplotlib",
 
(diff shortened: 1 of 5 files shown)


isaacrob-roboflow (approved):
I'd be interested in ap50 too since we report those in the benchmark, but leaving that up to you! nice clean interface here :)

PR #68, opened 2025-03-27
docs: 📝 RTSP Stream via Inference and OpenCV Stream examples added to README

onuralpszr:
## Docs

* Added a detailed example of using `InferencePipeline` with custom logic for processing RTSP streams, including a callback for predictions and a method for displaying annotated frames.

* Added an example of using OpenCV for capturing video from an RTSP stream, performing object detection with `RFDETRBase`, and displaying annotated frames.

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,6 +173,110 @@ cv2.destroyAllWindows()
 
 </details>
 
+<details>
+<summary>RTSP Stream</summary>
+
+## InferencePipeline from Inference Stream
+
+```python
+import cv2
+import numpy as np
+from inference import InferencePipeline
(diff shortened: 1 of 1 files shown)


onuralpszr:
cc @SkalskiP

probicheaux (approved):
looks good to me!

onuralpszr:
> looks good to me!

Thank you very much I tested both so hopefully will help others but Readme getting bigger maybe it is time to put a proper docs (Similar to inference,supervision) we can have more control over it. Also if you are okay I would like to assemble mkdocs for start as well

PR #70, opened 2025-03-27
Add W&B Metric Logging

mario-dg:
# Description

This pull request introduces a new logging feature to the `rfdetr` package, specifically adding support for Weights and Biases (W&B) logging. The most important changes include importing the necessary W&B module, adding a new class to handle W&B logging, and integrating this new logger into the training process. No depdency updates are needed, since `wandb` is already a dependency.

### New Feature: Weights and Biases Logging

* [`rfdetr/detr.py`](diffhunk://#diff-9f8117154976f0e714a38740eb63c7b864f2497955660e60fd2c83d8a848c814L15-R15): Imported `MetricsWandBSink` and added the `update` and `save` to the model callbacks.

## Type of change

A sample run that is logged to W&B can be seen below. The setup is similar to Tensorboard, whereas all 
`all_kwargs = {**model_config, **train_config, **kwargs, "num_classes": num_classes}` are saved as the `run.config`, as you can see in the second screenshot. This allows the user to compare different training runs, in terms of hyperparameters or specific model changes.
wandb-metrics
wandb-config

## Docs
Question: Should the README be updated, similar to the Tensorboard logging?

## More Questions
- Should the saved models files (`checkpoint.pth`, `checkpoint_best_ema.pth`, `checkpoint_best_regular.pth`, `checkpoint_best_total.pth`) also be uploaded as artifacts to the run at the end of training? Should this be configurable by the user?
- How about also uploading the `results.json` file and `metrics_plot.png` (created by `MetricsPlotSink`)?

### DIFF
--- a/README.md
+++ b/README.md
@@ -274,6 +274,35 @@ During training, two model checkpoints (the regular weights and an EMA-based set
   
 </details>
 
+### Logging with Weights and Biases (W&B)
+
+[W&B](https://www.wandb.ai) similar to TensorBoard is a very powerful platform to monitor training metrics. It is cloud-based and offers a more complex feature set compared to TensorBoard.
+
+<details>
+<summary>Get Started</summary>
+
+- To use W&B, make sure you are logged in:
+
+    ```bash
+    wandb login
+    ```
+
+    You will need to specify an API Key for your account that can be retrieved at https://wandb.ai/authorize.
+
+- For those familiar with W&B, you can specify the `project_name` and `run_name` in the `model.train` method:
+
+    ```python
+    from rfdetr import RFDETRBase
+
+    model = RFDETRBase()
+
+    detections = model.train(dataset_dir=<DATASET_PATH>, project_name=<PROJECT_NAME>, run_name=<RUN_NAME>)
+    ```
+
+    For those unfamiliar with W&B: projects are collections of related machine learning experiments, while runs are individual units of computation within a project that record specific experiments, such as training a model or conducting hyperparameter tuning. Not specifying a name, will result in random names provided by W&B.
+  
+</details>
+
 ### Load and run fine-tuned model
 
 ```python
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,7 +50,6 @@ dependencies = [
     "peft",
     "ninja",
     "einops",
-    "wandb",
     "pandas",
     "pylabel",
     "onnx",
@@ -69,7 +68,8 @@ onnxruntime = [
     "onnxruntime"
 ]
 metrics = [
-    "tensorboard>=2.13.0"
+    "tensorboard>=2.13.0",
+    "wandb"
 ]
 
 [project.urls]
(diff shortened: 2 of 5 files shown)


onuralpszr:
Hello @mario-dg I love the visuals and here is the couple the pointer. First of all package install wandb by default and that is something we should be able to avoid it. I notice that If pacakge not found similar to tensorboard we setting the none and silenting. There is also another metric in the code 

https://github.com/roboflow/rf-detr/blob/main/rfdetr/util/misc.py#L172C7-L172C19

If we modify this part just a little for start to make sure it won't asking for import and It gives us to move wandb pacakge to optional section. We also gain speed on install, less import load on library itself. What do you think @SkalskiP  ? 

That is my initial review based on what I see and understand. 

I hope you guys enjoy
Thank you.

mario-dg:
Hi @onuralpszr, thanks for the feedback. I'm with you in moving the W&B depdency into the optional metrics section of the `pyproject.toml` file. The already existing `MetricLogger` seems to have rough W&B support, but it's usage is disabled by default, and to enable it the user needs to modify the `train_one_epoch` and `evaluate` method in `engine.py.`

onuralpszr:
> Hi @onuralpszr, thanks for the feedback. I'm with you in moving the W&B depdency into the optional metrics section of the `pyproject.toml` file. The already existing `MetricLogger` seems to have rough W&B support, but it's usage is disabled by default, and to enable it the user needs to modify the `train_one_epoch` and `evaluate` method in `engine.py.`

Usage disabled but import is there so we need move that part and make sure it won't conflict and also import check so making sure it won't cause a problem.

mario-dg:
@onuralpszr, what do you think of this approach? The parameter renaming shouldn't result in any breaking changes and W&B is an optional dependency in the existing `MetricsLogger`.

SkalskiP:
Hey @probicheaux @isaacrob-roboflow — any objections if we nuke the current `MetricLogger`?

It’s becoming more of a pain than a help, especially since users have to modify both `train_one_epoch` and `evaluate` in `engine.py` just to customize logging. We’d like to standardize logging so it behaves more like the TensorBoard logger — simple, modular, and optional.

Another solid reason to get rid of it: [this issue](https://github.com/roboflow/rf-detr/issues/77). Turns out `wandb` is being imported during inference, which is clearly not ideal and is breaking things for some users.

Let me know if you're good with the change.

mario-dg:
Maybe we can provide a `MetricsBaseSink`, that users can override and add to the list of callbacks?

isaacrob-roboflow:
@SkalskiP go ahead and get rid of it!

SkalskiP:
@isaacrob-roboflow that's what I wanted to hear! 🔥

mario-dg:
Should this be done in a seperate PR or in this one?

probicheaux:
I'm in favor of merging this as is and nuking the MetricsLogger in a separate PR

probicheaux (approved):

probicheaux:
merged, @mario-dg thanks for your contribution!

PR #88, opened 2025-03-31
remove outdated lr scheduler state control for resume

isaacrob-roboflow:
# Description

We originally used a different learning rate scheduler framework but then switched to LambdaLR and forgot to update our resume training logic. This fixes that.

I am also removing a param group update during the resume. Probably there to make it easier to change learning rates in different parts of the model during resume, but it reeks of code smell to me so I'm taking it out as iirc we don't plan to support that anyway

## Type of change

YOUR_ANSWER

## Docs


probicheaux (approved):

SkalskiP:
I'm testing it now ;)

PR #76, opened 2025-03-28
Add C++ ONNX Demo

sctrueew:
This PR introduces a C++ demo for the RF-DETR model, allowing users to perform real-time object detection using an ONNX model. The demo supports various input sources, including images, videos, and live camera streams, with optional CUDA acceleration.

Key Features:
✅ Loads an RF-DETR model in ONNX format
✅ Supports image, video, and live camera inference
✅ Enables CPU and CUDA (GPU) execution
✅ Configurable confidence threshold for detections
✅ Outputs annotated images/videos with detected objects
✅ Uses COCO class labels for object recognition

Run Examples:
🔹 Image Inference:
Detect objects in a static image and save the output:

```
./main--model path/to/model.onnx --source_type image \
  --input path/to/image.jpg --output path/to/output.jpg \
  --conf 0.6 --labels path/to/coco.names
```

🔹 Video Inference:
Process a video file and save the annotated output:

```
./main --model path/to/model.onnx --source_type video \
  --input path/to/video.mp4 --output path/to/output.mp4 \
  --conf 0.5 --use_cuda
```
🔹 Live Camera Inference (Default ID 0):
Run inference on the default webcam (ID 0) with GPU acceleration:

```
./main --model path/to/model.onnx --source_type camera \
  --input 0 --conf 0.55 --use_cuda
```
🔹 Live Camera Inference (Specific Camera ID 1):
Run inference on a specific camera (ID 1):

```
./main --model path/to/model.onnx --source_type camera \
  --input 1 --conf 0.55
```
🔹 Get Help & Available Options:

`./main --help`
Dependencies:

- OpenCV (for image and video processing)
- ONNX Runtime (for model inference)
- CUDA (optional, for GPU acceleration)

PR #114, opened 2025-04-04
RF-DETR TensorRT C++ demo

mohamedsamirx:
# Description

This pull request introduces a C++ TensorRT demo for building the RF-DETR TensorRT engine and utilizing it in C++. The implementation includes support for CUDA preprocessing.
I have Put everything in the same directory as this https://github.com/roboflow/rf-detr/pull/76 pull request for faster review process

Tested it on a photo and a video and both work really well
frame_0

### DIFF
--- a/inference/cpp/tensorrt/CMakeLists.txt
+++ b/inference/cpp/tensorrt/CMakeLists.txt
@@ -1,0 +1,54 @@
+﻿cmake_minimum_required(VERSION 3.18)
+project(rfdetr-tensorrt LANGUAGES CXX CUDA)
+
+# Set C++ standard
+set(CMAKE_CXX_STANDARD 17)
+add_definitions(-DAPI_EXPORTS)
+
+# Add source files
+set(SOURCES
+    main.cpp
+    src/preprocess.cu
+)
+
+# Check for rfdetr.cpp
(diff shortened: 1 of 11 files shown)


isaacrob:
we're already planning to do tensorrt native in this repo, yes? @Borda do you have thoughts on the benefit of having c++ native support in this repo? I don't feel like I understand the value

mohamedsamirx:
I thought I would never hear about this PR again. It is understandable if the library focuses only on Python. However, C++ is usually for people looking for deployment. I don't know if this is part of the Roboflow team's plan, but that was my intention when opening it.

I would love to work on other TensorRT-related tasks if the library needs contributors.

isaacrob:
haha @mohamedsamirx yeah trying to clean up lingering PRs :) now that @Borda is onboard and making sure what would otherwise have just been a research repo is up to snuff for long term relevance 

the pipeline we've been suggesting for folks for deployment is via https://github.com/roboflow/inference .. what is the utility of having a devoted demo here instead of just providing folks with the tensorrt engine and having them use whatever else they'd normally use to deploy that within c++?

mohamedsamirx:
I mean yeah, your thoughts make sense

PR #119, opened 2025-04-05
class names property for easy access

farukalamai:
# Description

This PR adds a class_names property to the RFDETR class that dynamically provides access to class names in dictionary format. The property retrieves class names from the loaded model if available, otherwise it falls back to COCO_CLASSES. The class IDs in the dictionary start from 1, providing a consistent interface for class identification. I think this small changes will be mostly suitable for the custom model to get access the class_names with ids.

## Type of change

I have changed the main.py to extract the class_names from model args. And added a property in the detr.py file

Tested with pre-trained models and custom models.

## Docs

- README.md file is updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,9 +84,9 @@ import requests
 import supervision as sv
 from PIL import Image
 from rfdetr import RFDETRBase
-from rfdetr.util.coco_classes import COCO_CLASSES
 
 model = RFDETRBase()
+CLASS_NAMES = model.class_names
 
 url = "https://media.roboflow.com/notebooks/examples/dog-2.jpeg"
 
@@ -94,7 +94,7 @@ image = Image.open(io.BytesIO(requests.get(url).content))
 detections = model.predict(image, threshold=0.5)
 
 labels = [
-    f"{COCO_CLASSES[class_id]} {confidence:.2f}"
+    f"{CLASS_NAMES[class_id]} {confidence:.2f}"
     for class_id, confidence
     in zip(detections.class_id, detections.confidence)
 ]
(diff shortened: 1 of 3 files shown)


farukalamai:
Hey @SkalskiP do you have any suggestions for this PR?

SkalskiP:
Hi @farukalamai 👋🏻 I just tested your PR and in general it works, but I found one edge case when it doesn't. [colab](https://colab.research.google.com/drive/1cUmyqS3O9P1OccL0s3eWLjDQO1bXobjq?usp=sharing)

It works when you:

- load pre-trained model from checkpoint
- load fine-tuned-model from checkpoint

It does not work when you run inference with model that was just fine-tuned.

```python
from rfdetr import RFDETRBase

model = RFDETRBase()

model.train(dataset_dir=dataset.location, epochs=3, batch_size=16, grad_accum_steps=1, lr=1e-4)

model.class_names
# still COCO class names
```

@probicheaux @isaacrob-roboflow I think we should update the `class_names` of the model right when we update `num_classes`.

probicheaux:
class_names would be updated when we set num_classes for training, in detr.py

SkalskiP:
@probicheaux s this currently happening? because I don't se that behaviour here.

farukalamai:
@SkalskiP I checked the collab notebook. Maybe we should add the class_names in the config.py file's ModelConfig class. hey @probicheaux can you check what's the issue

SkalskiP:
Hi @farukalamai! I spoke with @probicheaux yesterday — looks like he’s identified the source of the issue. He’ll either commit a fix directly to your PR or merge it and open a follow-up PR to cover this edge case. Stay tuned!

probicheaux (approved):

PR #130, opened 2025-04-08
Save weights to custom folder of user 's choice

valavanisleonidas:
# Description                                                                                                                              
                                                                                                                                           
This PR enables users to control where pretrained model weights are cached by                                                              
introducing an `RF_HOME` environment variable (defaulting to `~/.roboflow/models`).
                                                                                                                                           
Set `RF_HOME` to redirect downloads to any directory:                                                                                      
                                                                                                                                           
```bash                                                 
export RF_HOME=/mnt/shared/models
```
                                                                                                                                           
Bare filenames (e.g. rf-detr-base.pth) are resolved relative to this cache                                                                 
directory instead of the current working directory. Paths that already contain a                                                           
directory component are used as-is (parent directories are created if absent).                                                             
                                                                                                                                           
Changes                                                                                                                                    
                                                                                                                                           
- get_model_cache_dir() added to rfdetr.assets — reads RF_HOME, defaults                                                                   
to ~/.roboflow/models         1              #    
- ModelConfig.expand_path field validator resolves bare filenames to the cache                                                             
dir for explicitly-provided pretrain_weights values                                                                                        
- RFDETR.maybe_download_pretrain_weights fallback handles Pydantic field                                                                   
defaults (not run through validators unless validate_default=True)                                                                         
                                                                                                                                           
Type of change                                                                                                                             
                                                                                                                                           
- New feature (non-breaking change which adds functionality)                                                                               
                              1              #    
How has this change been tested?                                                                                                           
                                                        
Unit tests cover:                                                                                                                          
- get_model_cache_dir() default + RF_HOME override
- Bare filename resolution via ModelConfig + maybe_download_pretrain_weights                                                               
- makedirs for both cache dir (bare filename) and explicit path directories 
- None pretrain_weights skips download


SkalskiP:
Hi @valavanisleonidas 👋🏻 Thank you so much for your interest in RF-DETR and for submitting this PR!

We won’t be merging this PR in its current form, and here’s why:

- No other major library handles this in quite the same way. For example, libraries like Transformers (Hugging Face) allow customization via a `cache_dir` argument or environment variables like `HF_HOME`. Ultralytics, on the other hand, doesn’t support custom paths at all and sticks to a default location.
- Currently, `pretrain_weights` should either be a valid path to existing weights on disk or one of the predefined model IDs for downloading to a local default location. In the future, we might extend it to support valid model IDs from the Roboflow platform as well.
- If we were to enable customization of the model cache directory, it would be better implemented through an additional argument like `cache_directory` or an environment variable such as `RF_HOME`, aligning with established practices and offering a cleaner, more consistent user experience.

Thanks again for your contribution! Let me know if you’d like to discuss this further or explore alternative approaches.

valavanisleonidas:
Hello @SkalskiP,

thank you for your response. I was trying to implement something with minimal changes to the code. To be honest I took the idea from Ultralytics in which you can do `YOLO("models/yolo12l.pt")` and it downloads the file in folder `models`. It is valid as long as you put the correct model name. 

Another idea is to have two properties one for selecting the model to download `rf-detr-base` , `rf-detr-base2` and `rf-detr-large` and one for the path to download it to. 

I just believe it's really useful to be able to select where to save the model if it is not in a `.cache` known folder somewhere because depending on the file you run, the folder level in the project structure you are in, you need to change the path or redownload it

SkalskiP:
Yes, I believe the responsibility should be split between the two arguments.

Borda:
@valavanisleonidas, apologies for being silent for quite some time. I wanted to quickly click with you if you are still interested in working on this PR, as the title suggests?

> To be honest I took the idea from Ultralytics in which you can do `YOLO("models/yolo12l.pt")` and it downloads the file in folder `models`. The path you input to `Yolo()` is valid as long as you put the correct model name (`yolo12l.pt` etc).

I don't think this is a good pattern to follow

> * If we were to enable customization of the model cache directory, it would be better implemented through an additional argument like `cache_directory` or an environment variable such as `RF_HOME`, aligning with established practices and offering a cleaner, more consistent user experience.

Yes, I agree with this one...

valavanisleonidas:
Hello @Borda, yes I can continue with the PR. the selected path is to add an env var `cache_directory` for example ?

Borda:
> Hello, yes I can continue with the PR. the selected path is to add an env var `cache_directory` for example ?

Personally, I would take the simple path for now, let's have env. variable what would specify cache/home directory, which by default can be, for example, `~/.roboflow/models`. What do you think?

valavanisleonidas:
ok 

> > Hello, yes I can continue with the PR. the selected path is to add an env var `cache_directory` for example ?
> 
> Personally, I would take the simple path for now, let's have env. variable what would specify cache/home directory, which by default can be, for example, `~/.roboflow/models`. What do you think?

Ok sounds good!

Borda:
> Ok sounds good!

Cool, are you going to continue? 🦝

valavanisleonidas:
> > Ok sounds good!
> 
> Cool, are you going to continue? 🦝

Hey i have been busy lately. i plan to continue and fix it this week :)

Borda (approved):

PR #140, opened 2025-04-10
[BUG] Predict Correct calculation for width, height 

valavanisleonidas:
# Description

There was a last minute change in predict batch inference which moved the calculation of the original image size after the resize and this leads to incorrect results

This code should be before the image resize
```
 h, w = img_tensor.shape[1:]
 orig_sizes.append((h, w))
```

## Type of change

Manually with an image to see I get correct results 

## Docs


valavanisleonidas:
@SkalskiP I did a last minute bug and for some reason didnt check it until now. This PR fixes it :D

SkalskiP (approved):

PR #155, opened 2025-04-13
support local-encoder-weights

rajatprasun:
# Description

This change adds support for **local encoder weights***, which is very similar to the existing **'pretrain_weights'** parameter. The **'pretrained_encoder'** parameter already exists in the backbone but is not exposed in the config.py.  This parameter is essential when the user is not able to download the weights from the huggingface hub in the target machine.  If this parameter is not specified, the default behavior will kick in, and it will try to download weights from huggingface. 

Users can download the weights separately and set this parameter:
```
model = RFDETRBase(pretrain_weights=pretrain_weights,pretrained_encoder=pretrained_encoder)
```

### DIFF
--- a/rfdetr/config.py
+++ b/rfdetr/config.py
@@ -12,6 +12,7 @@ DEVICE = "cuda" if torch.cuda.is_available() else "mps" if torch.backends.mps.is
 
 class ModelConfig(BaseModel):
     encoder: Literal["dinov2_windowed_small", "dinov2_windowed_base"]
+    pretrained_encoder:Optional[str] = None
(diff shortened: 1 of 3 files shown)


SkalskiP:
Hi @rajatprasun 👋🏻 thanks a lot for your interest in RF-DETR. I believe that change is no longer required. 

This week we merged https://github.com/roboflow/rf-detr/pull/136, effectively removing the call to HF altogether. Please try to install RF-DETR from source (`
pip install git+https://github.com/roboflow/rf-detr.git
`) and check if that solves your problem.

PR #158, opened 2025-04-14
replace print with logger calls + remove unused imports

SkalskiP:
# Description

This pull request includes a series of changes primarily focused on replacing print statements with logger calls.


rfdetr/datasets/coco.py
```diff
@@ -191,58 +194,55 @@ def make_coco_transforms_square_div_64(image_set, resolution, multi_scale=False,
...
+        
+        img_folder, ann_file = PATHS[image_set.split("_")[0]]
+        square_resize_div_64 = getattr(args, 'square_resize_div_64', False)
+        
+        if square_resize_div_64:
+            logger.info(f"Building Roboflow {image_set} dataset with square resize at resolution {resolution}")
+            dataset = CocoDetection(img_folder, ann_file, transforms=make_coco_transforms_square_div_64(image_set, resolution, multi_scale=args.multi_scale, expanded_scales=args.expanded_scales))
+        else:
+            logger.info(f"Building Roboflow {image_set} dataset at resolution {resolution}")
+            dataset = CocoDetection(img_folder, ann_file, transforms=make_coco_transforms(image_set, resolution, multi_scale=args.multi_scale, expanded_scales=args.expanded_scales))
```
isaacrob-roboflow, line 244: there is a bug in the existing implementation that is stopping expanded_scales from getting propagated to the dataset when training on roboflow data. you can see it clearly here. this change fixes that bug, which will likely increase performance. however, it'll also slow down training a bit and increase memory usage a bit. not sure that's what we want. worth discussing

rfdetr/engine.py
```diff
@@ -70,19 +72,25 @@ def train_one_epoch(
...
     else:
         scaler = GradScaler('cuda', enabled=args.amp)
 
     optimizer.zero_grad()
-    assert batch_size % args.grad_accum_steps == 0
+    
+    # Check if batch size is divisible by gradient accumulation steps
+    if batch_size % args.grad_accum_steps != 0:
+        logger.error(f"Batch size ({batch_size}) must be divisible by gradient accumulation steps ({args.grad_accum_steps})")
+        raise ValueError(f"Batch size ({batch_size}) must be divisible by gradient accumulation steps ({args.grad_accum_steps})")
```
isaacrob-roboflow, line 86: we should be able to handle this case but I am not sure it is being handled correctly right now. nothing fundamental stopping us

rfdetr/util/metrics.py
```diff
@@ -1,8 +1,16 @@
+# ------------------------------------------------------------------------
+# RF-DETR
+# Copyright (c) 2025 Roboflow. All Rights Reserved.
+# Licensed under the Apache License, Version 2.0 [see LICENSE for details]
+# ------------------------------------------------------------------------
```
isaacrob-roboflow, line 5: this file was probably also copied over from LW-DETR .. even though they don't have a trademark in theirs, should we add it for them? (assuming we did copy it .. )

isaacrob-roboflow (commented):
couple notes, need to discuss the expanded_scales bug fix

rfdetr/util/metrics.py
SkalskiP: @isaacrob-roboflow This wasn’t copied from LW-DETR — I created it myself. All the code here is either written by us or by open-source contributors.

rfdetr/engine.py
SkalskiP: @isaacrob-roboflow My PR didn’t change the behavior here — it just made it consistent with how the logger is used elsewhere. I’m happy to explore better approaches, but that’s outside the scope of this PR.

rfdetr/datasets/coco.py
SkalskiP: @isaacrob-roboflow what's your suggestion here?

isaacrob-roboflow: let's do both and tag the issue in the comment?

Borda (dismissed):

Borda:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/roboflow/rf-detr/pull/158#pullrequestreview-3780356444)

PR #159, opened 2025-04-14
Fix wandb import on export

Matvezy:
# Description

Fix wandb dependency to only be imported when requested

## Type of change

Tested locally

No

## Docs

-   [ ] Docs updated? What were the changes: No


Matvezy:
We don't need it, and we actually don't want it. Let's drop it then 👍

SkalskiP (approved):

Issue #66, opened 2025-03-27
tensor shapes not matching for custom trained model


Akhp888:
### Search before asking

- [x] I have searched the RF-DETR issues and found no similar bug report.

### Bug

Hello,

I get the below error when i run inference on the custom model I trained.
detections = model.predict(path)

RuntimeError: output with shape [1, 778, 1034] doesn't match the broadcast shape [3, 778, 1034]

### Environment

Name: rfdetr
Version: 1.0.8
Summary: RF-DETR

### Minimal Reproducible Example

detections = model.predict(path)

### Are you willing to submit a PR?

- [ ] Yes, I'd like to help by submitting a PR!

probicheaux:
Hi @Akhp888 , can you please provide more information about which image you're running it on, and how model was initialized? Is it possible your image is grayscale?

Akhp888:
The images are 3 channel not grayscale , both Trained and inferred .

The training was done by below command

```
from rfdetr import RFDETRBase

model = RFDETRBase()

model.train(dataset_dir="<path to directory>", epochs=50, batch_size=4, grad_accum_steps=4, lr=1e-4) 
```

Akhp888:
Based on the error I also tried to convert the image to greyscale before passing through predict but still get the same error .
Below is the full error

`---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
Cell In[11], line 10
      6 print(image.mode)
      8 model = RFDETRBase(pretrain_weights='/home/ec2-user/SageMaker/notebook/rf-detr/output/checkpoint_best_regular.pth')
---> 10 detections = model.predict(path)

File ~/anaconda3/envs/pytorch_p310/lib/python3.10/site-packages/rfdetr/detr.py:102, in RFDETR.predict(self, image_or_path, threshold, **kwargs)
     99     h, w = image_or_path.shape[1:]
    101 image = image.to(self.model.device)
--> 102 image = F.normalize(image, self.means, self.stds)
    103 image = F.resize(image, (self.model.resolution, self.model.resolution))
    105 predictions = self.model.model.forward(image[None, :])

File ~/anaconda3/envs/pytorch_p310/lib/python3.10/site-packages/torchvision/transforms/functional.py:349, in normalize(tensor, mean, std, inplace)
    346 if not isinstance(tensor, torch.Tensor):
    347     raise TypeError(f"img should be Tensor Image. Got {type(tensor)}")
[...]

PR #179, opened 2025-04-23
Refactor: raise early on invalid PIL input, removed else block

gaetanDubuc:
# Description

As mentionned in issue #66, the input shape was not being checked when a PIL image was given to the RFDETR.predict function. To address this, the else block has been removed, and now a grayscale image will raise the same error as a wrongly shaped torch tensor.

## Type of change

-   [X] Bug fix (non-breaking change which fixes an issue)

The following script was used for testing. I could add it to the PR, but I believe you need to first define clear unit test guidelines:

```python

import pytest
import torch
from PIL import Image

from .detr import  RFDETRBase


model = RFDETRBase()

@pytest.mark.parametrize("image", [
    torch.rand(3, 100, 100),
    Image.new("RGB", (100, 100)),
])
def test_predict_with_valid_tensor(image):
    detections = model.predict(image)
    assert detections is not None

def test_predict_with_pixels_above_one():
    image = torch.rand(3, 100, 100) * 255  # not normalized
    with pytest.raises(ValueError, match="Image has pixel values above 1"):
        model.predict(image)

@pytest.mark.parametrize("image", [
    torch.rand(1, 100, 100),
    Image.new("L", (100, 100)),
])
def test_predict_with_invalid_tensor_shape(image):
    with pytest.raises(ValueError, match="Invalid image shape"):
        model.predict(image)
```

There are no deployment considerations as no functionality change was made, just a bug fix related to input validation.

## Docs

No documentation was updated.

### DIFF
--- a/rfdetr/detr.py
+++ b/rfdetr/detr.py
@@ -170,20 +170,20 @@ class RFDETR:
                 img = Image.open(img)
 
             if not isinstance(img, torch.Tensor):
-                img_tensor = F.to_tensor(img)
-            else:
-                if (img > 1).any():
-                    raise ValueError(
-                        "Image has pixel values above 1. Please ensure the image is "
-                        "normalized (scaled to [0, 1])."
-                    )
-                if img.shape[0] != 3:
-                    raise ValueError(
-                        f"Invalid image shape. Expected 3 channels (RGB), but got "
-                        f"{img.shape[0]} channels."
-                    )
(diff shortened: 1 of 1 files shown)


SkalskiP (approved):

SkalskiP:
looks good! @gaetanDubuc thanks a lot for the PR. merging!

PR #214, opened 2025-05-16
Build "optimize_for_inference" method

isaacrob-roboflow:
# Description

RF-DETR is designed to be flexible, but that can make it slower at runtime, due to things like dynamic resize of the positional embedding. Those issues are solved when we export for compilation, so don't show up in our TRT report, but are felt during torch inference.

This PR uses a variety of techniques to speed up pytorch inference. Overall, on my L4 GPU I see a >2x end to end speedup (25ms to 12ms) by calling the optimize_for_inference method of RF-DETR prior to running inference. It stops the positional reinterpolation, compiles via jit trace, and supports native bfloat16 inference which is especially useful for large batch sizes. The jit trace takes <3 seconds for me so I think it is reasonable to have it on by default.

I also enable us to take advantage of tensor cores when available by setting float32 matmul precision to 'high'.

I also replace the duplicated MHA implementation with the pytorch-native one, which allows us to avoid directly computing attention weights which speeds us up and saves memory. I believe there is no change in training behavior with this change based on my experiments, but happy to discuss.

I also tried to ensure there were sufficient warnings and well-explained error modes so that this extra mode is easily understood.

For reference, I see YOLO11m from ultralytics getting 17ms end to end on the same image in the same script.

Cheers!

## Type of change

-   [ ] Bug fix (non-breaking change which fixes an issue)
-   [x] This change requires a documentation update

Tested locally with an inference script and a train script.

## Docs

-   [x] Docs updated? What were the changes:

I added a little bit saying the optimize_for_inference method is available. If you call .predict without having called it, the first time you do so you'll be warned, so it should be clear that the feature exists.


isaacrob-roboflow:
@SkalskiP @probicheaux please do not merge until I've had more time to verify that the MHA replacement doesn't do anything bad. but I would appreciate an approval in the meantime ;) :)

probicheaux (dismissed):

probicheaux (approved):

PR #239, opened 2025-06-02
do random resize on batched images

isaacrob-roboflow:
# Description

Right now we randomly resize each image independently and then pad the batch of images to have the same dimensions as the largest image. This means we waste compute and also that with probability ~1 the model doesn't see images of the target resolution without padding.

This PR moves the random resize to be directly in front of the forward pass. This gives slightly higher accuracy and slightly higher throughput.

It also fixes a bug where the expanded scales argument was ignored for Roboflow datasets. This should give slightly better results but slightly lower throughput.

## Type of change

Tested locally via a training script. 

## Docs

### DIFF
--- a/rfdetr/config.py
+++ b/rfdetr/config.py
@@ -76,6 +76,7 @@ class TrainConfig(BaseModel):
     output_dir: str = "output"
     multi_scale: bool = True
     expanded_scales: bool = True
+    do_random_resize_via_padding: bool = False
     use_ema: bool = True
     num_workers: int = 2
     weight_decay: float = 1e-4
--- a/rfdetr/datasets/coco.py
+++ b/rfdetr/datasets/coco.py
@@ -107,7 +107,7 @@ class ConvertCoco(object):
         return image, target
 
 
-def make_coco_transforms(image_set, resolution, multi_scale=False, expanded_scales=False):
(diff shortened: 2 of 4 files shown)


Borda:
Hey @isaacrob-roboflow — thanks for digging into this and putting together a clear write-up.

The problem you identified is real: per-image random resize followed by batch-padding does mean the model almost never sees a clean target-resolution crop, and the `expanded_scales` omission in `build_roboflow` was a genuine silent bug.

Unfortunately, this PR has been superseded by changes that were merged into `develop` while it was open. The batch-level resize logic (interpolating the full batch to a randomly chosen scale right before the forward pass) is now in `src/rfdetr/training/module_model.py`, and the `expanded_scales` fix for the Roboflow dataset path landed in `src/rfdetr/datasets/coco.py`. The `do_random_resize_via_padding` escape hatch is there too.

Because `develop` has also restructured the package layout from `rfdetr/` to `src/rfdetr/`, rebasing this branch would amount to re-implementing work that already exists — not worth the effort on your end.

I'm going to close this one. The direction you took here was the right call, and it's genuinely reflected in what landed. If you spot other inefficiencies in the training pipeline, we'd love to see more PRs.

PR #245, opened 2025-06-19
Soft-RFDETR : An optionnal Mixture of Experts Module to RFDETR

LeosCtrt:
# Description

Hello, I propose here an optional soft mixture of experts implementation for RF-DETR. The idea comes from the article [From Sparse to Soft Mixtures of Experts](https://arxiv.org/abs/2308.00951), and it seems to usually improve the overall performance of a model. I thought that it would be a nice option to add. 

As the article suggests, I set the default value of experts per slots to 1. 

In order to add this option we would need to add the python library "soft_moe". This library offers good a Soft MoE Wrapper, the source code is available [here](https://github.com/bwconrad/soft-moe/tree/main).

## Type of change

I tested it on a small private dataset and it improved a bit the accuracy (I gained roughly 1% mAP, so a small gain but still a gain). I am not able to test it on the full COCO dataset but I guess it would also improve a bit the overall accuracy. 

As expected the MoE RF-DETR model is slightly slower than the basic one, but still very fast.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -58,6 +58,7 @@ dependencies = [
     "pydantic",
     "supervision",
     "matplotlib",
+    "soft_moe",
 ]
 
 [project.optional-dependencies]
--- a/rfdetr/config.py
+++ b/rfdetr/config.py
@@ -31,7 +31,7 @@ class ModelConfig(BaseModel):
     group_detr: int = 13
     gradient_checkpointing: bool = False
     MoE: bool = False
-    MoE_params: List[int] = [16, 1]
+    MoE_params: List[int] = [32, 1]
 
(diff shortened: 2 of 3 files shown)


isaacrob-roboflow:
we are leveraging a pretrained backbone, so without robust pretrained weights we can't use a different architecture

isaacrob-roboflow:
additionally, have you looked at TRT speed for this? many of these types of architectures actually end up being meaningfully slower in terms of latency when compiled due to memory bottleneck

isaacrob:
going to close this PR in the name of cleaning up the repo. @LeosCtrt I would encourage you to try it on larger scale stuff and if you find it to be Pareto optimal, publish a paper :)

PR #246, opened 2025-06-23
[BUG][CLI] Fix error in cli script

surya3214:
# Description

Training with coco_dir option fails due to a small bug from L#52

`NameError: name 'device_supports_cuda' is not defined`

## Type of change

"python3 cli/main.py --coco_dir <path>" won't fail anymore due to "NameError" from "device_supports_cuda"

None

## Docs

None

### DIFF
--- a/rfdetr/cli/main.py
+++ b/rfdetr/cli/main.py
@@ -15,10 +15,8 @@ import torch
 from rf100vl import get_rf100vl_projects
 
 from rfdetr import RFDETRBase
-
-from rfdetr import RFDETRBase
 from rfdetr.config import DEVICE
-import os
+
 
 def download_dataset(rf_project: roboflow.Project, dataset_version: int):
     versions = rf_project.versions()


nok:
Hello @surya3214 , alternatively you can use the variable `DEVICE` from the [config.py#L11](https://github.com/roboflow/rf-detr/blob/develop/rfdetr/config.py#L11):

```python
def train_from_coco_dir(coco_dir: str):
    from rfdetr.config import DEVICE

    rf_detr = RFDETRBase()
    rf_detr.train(
        dataset_dir=coco_dir,
        epochs=1,
        device=DEVICE,
    )
```

Abdul-Mukit:
Why should it not be defined by the user?
e.g. In most CLI commands for DETRs we can usually set it by --device option.
```
parser.add_argument('--device', default='cuda',
                        help='device to use for training / testing')
```

PR #252, opened 2025-06-26
load best checkpoint for test metrics

isaacrob-roboflow:
# Description

The reported test metrics are not using the best weights, instead using the last weights. This is bad. This is fixed in this PR.

Also fixes a bug where run_test is not defined by default. Now it is both defined and we generate test stats.

## Type of change

Tested locally.

## Docs


probicheaux (approved):
lgtm

PR #331, opened 2025-08-21
Improve dataset handling with adaptive folder detection

yyq19990828:
## Summary

This PR improves the dataset handling functionality in `rfdetr/datasets/coco.py` by adding adaptive folder detection and better fallback mechanisms for test datasets.

### Key improvements:
- **Adaptive val/valid folder detection**: Automatically detects whether the dataset uses `val` (YOLO format) or `valid` (COCO format) folder naming
- **Smart test dataset fallback**: When test dataset is missing (which is common in most datasets), automatically uses the validation dataset as fallback
- **Reduced manual dataset modification**: Users no longer need to manually rename folders to match expected naming conventions
- **Better error handling**: Proper FileNotFoundError when neither val nor valid folders exist
- **Code internationalization**: Converted Chinese comments to English

### Why this change?
- Most datasets don't include a separate test split, making the fallback mechanism necessary
- YOLO datasets typically use `val` folder while COCO datasets use `valid` folder
- This reduces the friction for users when working with different dataset formats without requiring manual folder restructuring

## Test plan
- [x] Verify the code handles both `val` and `valid` folder structures
- [x] Confirm test dataset fallback works when test folder is missing
- [x] Ensure proper error handling when validation folders are missing

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/rfdetr/datasets/coco.py
+++ b/rfdetr/datasets/coco.py
@@ -238,11 +238,28 @@ def build_roboflow(image_set, args, resolution):
     root = Path(args.dataset_dir)
     assert root.exists(), f'provided Roboflow path {root} does not exist'
     mode = 'instances'
+    
+    # Adaptively detect val/valid folder name
+    val_folder = None
+    if (root / "valid").exists():
+        val_folder = "valid"
+    elif (root / "val").exists():
+        val_folder = "val"
+    else:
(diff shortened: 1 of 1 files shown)


Borda (commented):
We have added YOLO dataset support in `1.4`, so is this still needed?

Issue #330, opened 2025-08-20
train_from_config initializes one less class_embed neuron when training on custom 1-indexed dataset.


Abdul-Mukit:
### Search before asking

- [x] I have searched the RF-DETR issues and found no similar bug report.

### Bug

Similar to https://github.com/roboflow/rf-detr/issues/51  
When trying to train on a 1-indexed COCO dataset, training fails. 

`train_from_config()` has a bug.  

The `self.model.reinitialize_detection_head(num_classes)` call should have been self.model.reinitialize_detection_head(num_classes + 1)` instead.   
I think this originates as the name num_classes is quite misleading.
https://github.com/roboflow/rf-detr/blob/24ce179cb5d71d9049724c9f3bc25b506d9f42a4/rfdetr/models/lwdetr.py#L597

My references:
The original DETR's intent of doing num_classes + 1 is explained very well already.
https://github.com/facebookresearch/detr/issues/108#issuecomment-674854977
https://github.com/facebookresearch/detr/issues/108#issuecomment-650269223

Proposed Solution:
The following solution works for both 0-indexed and 1-indexed COCO datasets:
Even works with the original COCO datasets with "holes" in labels.

```
    def train_from_config(self, config: TrainConfig, **kwargs):
        with open(
            os.path.join(config.dataset_dir, "train", "_annotations.coco.json"), "r"
        ) as f:
            anns = json.load(f)
            # num_classes = len(anns["categories"])
            num_classes = max([category['id'] for category in anns["categories"]])
            class_names = [c["name"] for c in anns["categories"] if c["supercategory"] != "none"]
            self.model.class_names = class_names

        if self.model_config.num_classes != num_classes:
            logger.warning(
                f"num_classes mismatch: model has {self.model_config.num_classes} classes, but your dataset has {num_classes} classes\n"
                f"reinitializing your detection head with {num_classes} classes."
            )
            self.model.reinitialize_detection_head(num_classes + 1)
```

Note that when we initially define `self.class_embed` in LWDETR we actually pass:
https://github.com/roboflow/rf-detr/blob/24ce179cb5d71d9049724c9f3bc25b506d9f42a4/rfdetr/models/lwdetr.py#L605

But when we call `train_from_config` when don't do the same `num_classes + 1`.

[...]

Issue #51, opened 2025-03-25
num_classes issue


virgile-fsr:
https://github.com/roboflow/rf-detr/blob/6ca1b58e43668321c4eeea34135bb231cd30520a/rfdetr/main.py#L92

If user creates a new model with:
`model = RFDETRBase(num_classes=123)`
The code linked above will overwrite the num classes specified by the users by the one in the checkpoint (90 by default). I don't think this is expected behavior. More worrisome, the config of the object stays with num_classes=123 which will cause issues in the train function.

If you call: `model.train(dataset_dir=...)` with the model above, [it will check the number of classes in the dataset and compare it with self.model_config.num_classes](https://github.com/roboflow/rf-detr/blob/6ca1b58e43668321c4eeea34135bb231cd30520a/rfdetr/detr.py#L47) which is still 123 in our example (even though the actual model has a 90 sized output layer). So if the number of classes in the dataset is indeed 123, the check will pass even though it should not.

A different issue it that if you try a dataset with a single class, the labels from the dataloader will have size zero.

isaacrob-roboflow:
single class is being discussed here https://github.com/roboflow/rf-detr/issues/48, the issue is due to our working mainly with datasets exported from roboflow which have a filler 0 class causing the issue you're observing with non-roboflow format datasets. 

good point re using number of classes as an excuse to reinitialize the head! do you have an idea of what a preferred interface would look like? maybe something like having num_classes default to None and if it is not None, we reinitialize the head with that number of classes?

Akhp888:
Hello,
Glad that this issue is raised .I am facing the same issue when training on a dataset with 6 classes. Irrespective of me declaring the num of classes to 6 its reinitializing detection head with 90 classes. Is there a workaround for this?
Because as it seems now there is no possibility to train on custom data !

anyaflya:
I encounter the same issue. I want the model to predict two classes, and it still trained with 90 classes because of the reasons mentioned by virgile. Moreover, when predicting with the trained 90-class model 
`model = RFDETRBase(pretrain_weights="checkpoint_best_total.pth", num_classes=2)`
I have: 
`num_classes mismatch: pretrain weights has 1 classes, but your model has 2 classes`.

On my side, a preferred interface would be to have the model trained with default `num_classes` equal to the number of categories in the annotations coco json files.

panagiotamoraiti:
[...]

PR #356, opened 2025-09-12
Fix: incorrect num_classes used in RFDETR.train_from_config

timnielen:
# Description

TL;DR
The current framework handles the number of classes in a dataset incorrectly as discussed in the following issues:
https://github.com/roboflow/rf-detr/issues/330
https://github.com/roboflow/rf-detr/issues/51

In Detail:
The rf-detr framework calculates the number of classes as the number of classes in the dataset and reinitializes the head with this number in case it doesn't match the model's num_classes.
There are several issues with this:

First of all, the head should be of size `num_classes+1` to account for the fact that COCO datasets are 1 indexed. In the original DETR the head was even of size 92 because they add another +1 for the "no-object" class at the very end. This has been discussed in https://github.com/facebookresearch/detr/issues/108. There they note that in theory this is not necessary to do this because the logit with index 0 could also act as the "no-object" class. This is exactly why RF-DETR's head has a size of 91, meaning the 0 index class acts as the "no-object" class. (As far as I understand this is no problem as since Deformable DETR they stopped using a special weight in the cross_entropy loss for the "no-object" class which was at index -1 for DETR.)
**So why does this not cause any problems for datasets downloaded from ROBOFLOW?**
In roboflow datasets indexing starts at 0 and the 0th class is in fact no actual class it is the superclass of all other classes.
Here an example from the soccer players.v2 dataset:
```
[
        {
            "id": 0,
            "name": "soccer-players",
            "supercategory": "none"
        },
        {
            "id": 1,
            "name": "football",
            "supercategory": "soccer-players"
        },
        {
            "id": 2,
            "name": "player",
            "supercategory": "soccer-players"
        },
        {
            "id": 3,
            "name": "referee",
            "supercategory": "soccer-players"
        }
]
```
If you inspect it on ROBOFLOW https://universe.roboflow.com/roboflow-100/soccer-players-5fuqs it indeed only has 3 classes and not 4 so the first class is really just a dummy class. This is why by accident the current implementation works as `len(anns["categories"]) = 4 = number_of_actual_classes+1`
**However,** this will not work for any 1indexed dataset such as the original COCO dataset as it doesn't have this dummy class in the beginning.

Another issue is that `num_classes` should not be the number of classes in the dataset but rather the `max_id` because there might be some "holes". This is in fact the case for the original COCO dataset. It actually has only **80** classes in total and not 90 which is it's `max_id` because for some reason some indices are missing. 

I can only suspect the original RF-DETR models haven't been trained with this pipeline because this would have caused the head to be reinitialized with 80 instead of 91 logits as this is the `len(anns["categories"])` in the case of COCO.

The fix accounts for both issues and **will still work with ROBOFLOW datasets**. I set the num_classes to the `max_id` which is 3 in the above example. Then I reinitialize the head with size `num_classes+1` to account for both the "no-object" logit in the 0th index and the fact that labels range from 1 to `num_classes`.

Edit: I just noticed this is the same fix as proposed by https://github.com/roboflow/rf-detr/issues/330

## Type of change

Here is how I veryfied the problem:

I downloaded the soccer players.v2 dataset from roboflow as mentioned above. 
    
As the first category is only a supercategory there are only 3 actual categories. 
**Roboflow also states that there are 3 categories** https://universe.roboflow.com/roboflow-100/soccer-players-5fuqs

When I train the model on the dataset as is:
  ```
model = RFDETRNano()

  model.train(
      dataset_dir="soccer players.v2-release.coco"
      )
```

It reinitializes the head to size 4 which is correct as 4=3+1=num_classes+1 but only because of the dummy class in the dataset.

Now, to test what happens if I have a 1indexed dataset I removed the dummy class. Now it reinitializes it to only 3 logits and I get an "index out of bounds" error as expected as the "referee" category with id 3 coresponds to the 3rd index which doesn't exist.

After the changes this is handeled correctly and the head get initialized with 4 logits and it prints the correct number of classes (3) in the reinitialization warning.

### DIFF
--- a/rfdetr/detr.py
+++ b/rfdetr/detr.py
@@ -126,8 +126,8 @@ class RFDETR:
             os.path.join(config.dataset_dir, "train", "_annotations.coco.json"), "r"
         ) as f:
             anns = json.load(f)
-            num_classes = len(anns["categories"])
             class_names = [c["name"] for c in anns["categories"] if c["supercategory"] != "none"]
+            num_classes = len(class_names)
             self.model.class_names = class_names
 
         if self.model_config.num_classes != num_classes:
@@ -135,7 +135,7 @@ class RFDETR:
                 f"num_classes mismatch: model has {self.model_config.num_classes} classes, but your dataset has {num_classes} classes\n"
                 f"reinitializing your detection head with {num_classes} classes."
             )
-            self.model.reinitialize_detection_head(num_classes)
+            self.model.reinitialize_detection_head(num_classes+1)
         
         train_config = config.dict()
         model_config = self.model_config.dict()
--- a/rfdetr/detr.py
+++ b/rfdetr/detr.py
@@ -127,7 +127,7 @@ class RFDETR:
         ) as f:
             anns = json.load(f)
             class_names = [c["name"] for c in anns["categories"] if c["supercategory"] != "none"]
-            num_classes = len(class_names)
+            num_classes = max(c["id"] for c in anns["categories"] if c["supercategory"] != "none")
             self.model.class_names = class_names
 
         if self.model_config.num_classes != num_classes:


Borda (approved):

Borda:
It seem this was addressed in #588
Anyway, thank you for your help!

PR #399, opened 2025-10-08
Correctly update `num_select` during seg model finetuning

sungchul2:
# Description

When fine-tuning seg-preview model, `num_select` seems to be set to 200, but it is set to 300 like other detection models.
It is due to that this parameter isn't set in `SegmentationTrainConfig`, so it isn't updated correctly.
With this fix, `num_select` is correctly updated to 200.
| before | after |
| -- | -- |
| `Namespace(..., num_select=300, ...)` | `Namespace(..., num_select=200, ...)` |

## Type of change

As adding `num_select` parameter in `SegmentationTrainConfig`, the argument in `Namespace` is correctly updated like above.

NA

## Docs


sungchul2:

PR #429, opened 2025-10-30
bug: fixed bug where predictions was overwritten loosing data

AnonymDevOSS:
Fixes a logic error in predict() where tuple outputs from the model were converted to a dict before checking their length. As a result, the segmentation branch (len==3) was never reached and pred_masks were dropped.

## Type of change

-   [X] Bug fix (non-breaking change which fixes an issue)


Borda:
seems this was already resolved, but that you for your help 👍

Issue #446, opened 2025-11-11
RuntimeError: Index put requires the source and destination dtypes match, got BFloat16 for the destination and Float for the source.


kawabe-jiw:
### Search before asking

- [x] I have searched the RF-DETR issues and found no similar bug report.

### Bug

An error occurred when running training with `use_position_supervised_loss` set to True. This issue arises when training on an NVIDIA L4.

~~~bash
File "/home/ubuntu/rf_deter_train/rf_detr_train/train.py", line 102, in train
    model.train(**train_params)
  File "/home/ubuntu/.cache/pypoetry/virtualenvs/rf-detr-train-vhWH2Pc8-py3.10/lib/python3.10/site-packages/rfdetr/detr.py", line 83, in train
    self.train_from_config(config, **kwargs)
  File "/home/ubuntu/.cache/pypoetry/virtualenvs/rf-detr-train-vhWH2Pc8-py3.10/lib/python3.10/site-packages/rfdetr/detr.py", line 191, in train_from_config
    self.model.train(
  File "/home/ubuntu/.cache/pypoetry/virtualenvs/rf-detr-train-vhWH2Pc8-py3.10/lib/python3.10/site-packages/rfdetr/main.py", line 341, in train
    train_stats = train_one_epoch(
  File "/home/ubuntu/.cache/pypoetry/virtualenvs/rf-detr-train-vhWH2Pc8-py3.10/lib/python3.10/site-packages/rfdetr/engine.py", line 130, in train_one_epoch
    loss_dict = criterion(outputs, new_targets)
  File "/home/ubuntu/.cache/pypoetry/virtualenvs/rf-detr-train-vhWH2Pc8-py3.10/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1775, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/home/ubuntu/.cache/pypoetry/virtualenvs/rf-detr-train-vhWH2Pc8-py3.10/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1786, in _call_impl
    return forward_call(*args, **kwargs)
  File "/home/ubuntu/.cache/pypoetry/virtualenvs/rf-detr-train-vhWH2Pc8-py3.10/lib/python3.10/site-packages/rfdetr/models/lwdetr.py", line 551, in forward
    losses.update(self.get_loss(loss, outputs, targets, indices, num_boxes))
[...]

PR #447, opened 2025-11-11
Fix for the dtype mismatch error that occurs when use_position_supervised_loss=True is enabled.

kawabe-jiw:
This patch resolves a dtype mismatch error between BFloat16 and Float that occurs during training when use_position_supervised_loss=True is enabled. This fix addresses issue https://github.com/roboflow/rf-detr/issues/446.


isaacrob:
I'd rather just drop support for the loss. We've never used it, residual from LW-DETR

Borda (dismissed):

PR #451, opened 2025-11-12
Introduce background_class83422 so that class names index from 1

SolomonLake:
# Description

Introduce background_class83422 so that class_names index from 1, keeping in line with COCO standard format for class names.

## Type of change

Tested in notebook with new training.

none

## Docs

### DIFF
--- a/rfdetr/detr.py
+++ b/rfdetr/detr.py
@@ -130,7 +130,9 @@ class RFDETR:
             ) as f:
                 anns = json.load(f)
                 num_classes = len(anns["categories"])
-                class_names = [c["name"] for c in anns["categories"] if c["supercategory"] != "none"]
+                class_names = [c["name"] for c in anns["categories"] if c["supercategory"] != "none" and c["name"] != "background_class83422"]
+                class_names.insert(0, "background_class83422")
+                num_classes = len(class_names)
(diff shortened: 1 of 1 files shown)


Abdul-Mukit:
@SolomonLake please see this:
https://github.com/roboflow/rf-detr/issues/330#issuecomment-3497965395  

Original COCO style expects category-id 0 as background. All major CV libraries like Roboflow, Ultralytics, and Detectron2 are getting rid of the background class. All expect the category index to start from 0. Due to this consensus, it is easy to use/swap models trained from different sources. We know the exported model will always produce 0 for a class0. I tried to stick to a similar solution as you proposed, and it caused more pain on both the data management side and in deployment side.  

Please consider the solution from Detectron2. It can handle COCO dataset starting from 0, 1, to whatever. Can handle "holes" in categories as we have in COCO dataset.    
https://github.com/facebookresearch/detectron2/blob/a9c0821a12ad353fb2a96f019515990d5460c5ac/detectron2/data/datasets/coco.py#L102

https://github.com/facebookresearch/detectron2/blob/a9c0821a12ad353fb2a96f019515990d5460c5ac/detectron2/evaluation/coco_evaluation.py#L237

PR #476, opened 2025-11-29
Fix inverted probability logic in RandomExpand.__call__

pantment:
This PR fixes a bug in RandomExpand.__call__ where the probability check was inverted.
According to the docstring, prob is the probability to expand, but the previous implementation expanded with probability 1 - prob.
The condition has been corrected so that expansion now happens with the intended probability.

### DIFF
--- a/rfdetr/datasets/transforms.py
+++ b/rfdetr/datasets/transforms.py
@@ -391,7 +391,7 @@ class RandomExpand(object):
         self.fill_value = fill_value
 
     def __call__(self, img, target):
-        if np.random.uniform(0., 1.) < self.prob:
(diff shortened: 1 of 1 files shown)


Borda:
@pantment, it seems I cannot resolve the collision for you...

Borda:
Thanks for the catch, @pantment — and apologies for not reviewing this sooner. The augmentation pipeline has seen updates in v1.6 (including #863 fixing AUG_AGGRESSIVE). We'll verify whether the RandomExpand probability inversion still exists in the current code. Since this PR has accumulated conflicts, we're closing it — if the bug persists we'll address it in a dedicated fix. Appreciate the contribution!

PR #508, opened 2025-12-21
fix: correct ONNX export output names for segmentation models

lixiaolei1982:
- Modified export() method in rfdetr/main.py
- Added dynamic output name configuration
- Segmentation models now include 'masks' output
- Backward compatible with detection and backbone models

# Description

Fix an inconsistency in the ONNX export function where segmentation models were not correctly configured with three output names.

## Issue
When exporting segmentation models to ONNX format, the third output (`masks`) is not properly named in the output specification, which could lead to:
- Missing mask outputs in the exported ONNX model
- Potential export errors or warnings
- Incorrect behavior when loading the model in inference engines

## Changes Made
1. Modified the output name configuration logic in `export()` method
2. Added proper model type detection using `self.args.segmentation_head`
3. Dynamic output names based on model type:
   - Backbone-only: `['features']`
   - Segmentation: `['dets', 'labels', 'masks']`
   - Detection: `['dets', 'labels']`
  
## Type of change
-   [ ] Bug fix (non-breaking change which fixes an issue)


Borda:
It was already fixed in #402 🦩, but I really appreciate your contribution!

PR #167, opened 2025-04-16
Enhance WindowedDinov2 input validation and fix tensor reshaping

developer0hye:
# Description

Harden WindowedDinov2 input validation: replace the strippable `assert` with an explicit `raise ValueError` so the spatial-dimension check is always active (even under Python `-O`). No tensor reshaping changes are included.

## Type of change

Please delete options that are not relevant.

-   [ ] Bug fix (non-breaking change which fixes an issue)
-   [x] New feature (non-breaking change which adds functionality)
-   [ ] This change requires a documentation update

## How has this change been tested, please provide a testcase or example of how you tested the change?

```python
import io
import requests
import supervision as sv
from PIL import Image
from rfdetr import RFDETRBase

model = RFDETRBase(resolution=546)
CLASS_NAMES = model.class_names

url = "https://media.roboflow.com/notebooks/examples/dog-2.jpeg"

image = Image.open(io.BytesIO(requests.get(url).content))
detections = model.predict(image, threshold=0.5)

labels = [
    f"{CLASS_NAMES[class_id]} {confidence:.2f}"
    for class_id, confidence
    in zip(detections.class_id, detections.confidence)
]

annotated_image = image.copy()
annotated_image = sv.BoxAnnotator().annotate(annotated_image, detections)
annotated_image = sv.LabelAnnotator().annotate(annotated_image, detections, labels)

sv.plot_image(annotated_image)
```
image

## Docs

-   [ ] Docs updated? What were the changes:

PR #523, opened 2026-01-03
fix: correct off-by-one error in windowed DINOv2 early stopping

lixiaolei1982:
Fixed an off-by-one error in the early stopping logic of WindowedDinov2WithRegistersEncoder.forward().

Bug: The condition if i > int(self.config.out_features[-1][5:]): incorrectly compares a 0-based layer index (i) with a 1-based stage number.

Fix: Changed to if i > int(self.config.out_features[-1][5:]) - 1: to properly convert the stage number to a 0-based index before comparison.

Example:
If out_features[-1] = "stage8", stage number = 8
Corresponding layer index = 7 (0-based)
Old condition: i > 8 → never true for i in [0, 11]
New condition: i > 7 → stops when i reaches 8

Impact: The encoder now correctly stops after processing the last required transformer layer, improving computational efficiency.

### DIFF
--- a/rfdetr/models/backbone/dinov2_with_windowed_attn.py
+++ b/rfdetr/models/backbone/dinov2_with_windowed_attn.py
@@ -671,7 +671,7 @@ class WindowedDinov2WithRegistersEncoder(nn.Module):
             if output_hidden_states:
                 all_hidden_states = all_hidden_states + (hidden_states,)
             
-            if i > int(self.config.out_features[-1][5:]):
+            if i > int(self.config.out_features[-1][5:])-1:
                 # early stop if we have reached the last output feature
                 break
             


Borda:
Thanks for catching this, @lixiaolei1982 — and apologies for letting this sit without a response for so long. The DINOv2 windowed encoder was reworked substantially in v1.6.0 (transformers v5 migration, update_drop_path fixes), and #167 (merged to develop, coming in the next release) added further input validation and tensor reshaping fixes. We believe the off-by-one you identified has been addressed as part of that work. Since this PR has accumulated merge conflicts and the underlying code has changed, we're closing it. We really appreciate your contribution — please take a look at the current develop branch and let us know if you spot any remaining issues!

Issue #260, opened 2025-07-11
Test set required even if run_test is False


RhysDeLoach:
### Search before asking

- [x] I have searched the RF-DETR issues and found no similar bug report.

### Bug

Hey y'all,

The functionality to perform a test run once the model finishes training has been added to the develop branch. I’m not sure if this was intentional, but in implementing this feature, a requirement for a test set was introduced, even when run_test is set to False. The error below illustrates this issue:

FileNotFoundError: [Errno 2] No such file or directory: 'dataset/test/_annotations.coco.json'

This seems to occur because dataset_test is being built and data_loader_test created regardless of the run_test flag’s value. According to the traceback, the failure happens at line 196 of main.py (see below):

dataset_test = build_dataset(image_set='test', args=args, resolution=args.resolution)

Also, if run_test is going to default to True, I believe there should be more graceful handling to inform the user that they need to either provide a test set or set run_test to False.

Let me know what y'all think. 

Best,
Rhys

### Environment

RF-DETR: develop

### Minimal Reproducible Example

1. Download dataset with just a training and validation split.
2. Train model with run_test = False

### Are you willing to submit a PR?

- [x] Yes, I'd like to help by submitting a PR!

Borda:
Hi @RhysDeLoach — apologies for the delayed close. This was fixed by #628, which is included in v1.6.1. Training with `run_test=False` no longer requires a test split. Please upgrade and retry. Closing as fixed.

PR #628, opened 2026-02-03
Allow training on custom datasets without test splits when run_test=False

AHolliday:
## What does this PR do?

Modified main.py so that if the user passes run_test=False, the test-split dataset and dataloader are not created in Model.train().  This way, the user is not required to define a test split in their dataset if they don't intend to use it.

**Related Issue(s):** Fixes issue https://github.com/roboflow/rf-detr/issues/260

## Type of Change

## Testing

- [ x ] I have tested this change locally
- [    ] I have added/updated tests for this change

**Test details:**
There is currently no test set for main.py, so I have not added any tests.

## Checklist

- [ x ] My code follows the style guidelines of this project
- [ x ] I have performed a self-review of my own code
- [ x ] I have commented my code where necessary, particularly in hard-to-understand areas
- [ x ] My changes generate no new warnings or errors
- [ x ] I have updated the documentation accordingly (if applicable)


Vishalkagade (commented):
Confirmed this bug — when run_test=False is set but no test/ split exists in the dataset, training crashes with FileNotFoundError because dataset_test, sampler_test, data_loader_test, and base_ds_test were all built unconditionally. The run_test flag was only checked at evaluation time (line 500), not at dataset loading time. This fix correctly guards all test-related code paths.

Vishalkagade (approved):

Borda (dismissed):

PR #647, opened 2026-02-07
Fix/postprocess memory oom

lixiaolei1982:
# 提交更改（使用规范的提交信息）
git commit -m "fix: resolve CUDA OOM in postprocess with memory-optimized solution

Problem:
- Original PostProcess causes CUDA out-of-memory when handling sparse_forward() outputs
- Memory explosion during boolean mask conversion on GPU
- Incompatible output formats between forward() and sparse_forward()

Solution:
- Implement OptimizedPostProcess class with dual-path processing
- Add batch-wise mask processing with configurable batch size
- Enable early CPU offloading to reduce GPU memory pressure
- Support automatic mixed precision for further memory optimization
- Maintain full backward compatibility with existing code

Key Features:
- Handles both dense tensor outputs and sparse dictionary outputs
- Configurable mask_batch_size for memory control
- Automatic GPU memory cleanup between operations
- No breaking changes to existing interfaces

Testing:
- Resolves OOM on 11GB GPU with 1024x1024 masks
- Compatible with both forward() and sparse_forward() methods
- Maintains identical output format for downstream compatibility"


Borda (changes requested):
Thank you for taking care of this bug, but seems you changed too many unrelated files, please keep the PR focused...

I believe that the fix is in 53fe17e, so let's sync your develop branch with upstream and rebase the one commit

Issue #648, opened 2026-02-07
Training crashes after first epoch while training on YOLO dataset


forest-spells:
### Search before asking

- [x] I have searched the RF-DETR issues and found no similar bug report.

### Bug

When training with a YOLO dataset, the training crashes after the first epoch with:

```
Traceback (most recent call last):
  File "/mnt/ssd/184/PycharmProjects/thermal_fer/Scripts/Face_detection/Train_with_Roboflow.py", line 82, in <module>
    model.train(
  File "/mnt/ssd/184/anaconda3/envs/thermal_fer/lib/python3.11/site-packages/rfdetr/detr.py", line 94, in train
    self.train_from_config(config, **kwargs)
  File "/mnt/ssd/184/anaconda3/envs/thermal_fer/lib/python3.11/site-packages/rfdetr/detr.py", line 235, in train_from_config
    self.model.train(
  File "/mnt/ssd/184/anaconda3/envs/thermal_fer/lib/python3.11/site-packages/rfdetr/main.py", line 407, in train
    test_stats, coco_evaluator = evaluate(
                                 ^^^^^^^^^
  File "/mnt/ssd/184/anaconda3/envs/thermal_fer/lib/python3.11/site-packages/rfdetr/engine.py", line 431, in evaluate
    coco_evaluator.update(res)
  File "/mnt/ssd/184/anaconda3/envs/thermal_fer/lib/python3.11/site-packages/rfdetr/datasets/coco_eval.py", line 63, in update
    coco_dt = COCO.loadRes(self.coco_gt, results) if results else COCO()
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/ssd/184/anaconda3/envs/thermal_fer/lib/python3.11/site-packages/pycocotools/coco.py", line 314, in loadRes
    res.dataset['info'] = copy.deepcopy(self.dataset['info'])
                                        ~~~~~~~~~~~~^^^^^^^^
KeyError: 'info'
```

### Environment

- RF-DETR 1.4.1 
- python 3.10.12
- Ubuntu 22.04.5
- PyTorch 2.8.0+cu128
- A100 80GB

### Minimal Reproducible Example

Training crashed before proceeding to evaluation, at the line:
```py
model.train(
    dataset_dir=image_path,
    epochs=num_epochs,
    batch_size=batch_size,
    grad_accum_steps=grad_accum_steps,
    lr=lr,
    output_dir=ckpt_save_path,
[...]

PR #681, opened 2026-02-13
Fix: Missing 'info' key in COCO export format

francescobodria:
> The current implementation of the dataset export returns a dictionary missing the "info" key. According to the COCO dataset format specification, the "info" section is expected by many downstream tools and validation scripts. Adding this key ensures better compatibility with the COCO API and prevents potential KeyErrors during evaluation.


francescobodria:
#648 related

Borda (dismissed):

PR #687, opened 2026-02-14
Add `device` field to `TrainConfig`

juan-cobos:
`TrainConfig` (and its subclass `SegmentationTrainConfig`) lacked an explicit `device` field. This meant the training device was only reaching `populate_args` indirectly through `ModelConfig`'s dict merge in `train_from_config`, and any `device=` kwarg passed to `.train()` would flow through unvalidated `**kwargs`.

## What does this PR do?

Adding the field provides validated device selection at training time, consistent with `ModelConfig` which already has `device`, and proper support for `SegmentationTrainConfig` which inherits it automatically. It enable passing `device` argument when training `RFDETRSeg` models.

## Type of Change

## Testing

- I have tested this change locally


Borda:
Well, I think that device setting shall be part of the training configuration...

Borda:
@juan-cobos, why would you not use auto-device?

juan-cobos:
@Borda I guess it's mostly about being explicit with the device. If I build on top of the model, I have to set device any way. Moreover, to be consistent with RFDetr config which accepts device as an argument.

Borda (dismissed):

src/rfdetr/config.py
```diff
@@ -274,6 +274,7 @@ class TrainConfig(BaseModel):
     lr: float = 1e-4
     lr_encoder: float = 1.5e-4
     batch_size: int = 4
+    device: Literal["cpu", "cuda", "mps"] = DEVICE
```
Borda:
```suggestion
    device: Literal["auto", "cpu", "cuda", "mps"] = DEVICE
```

Issue #709, opened 2026-02-20
RF-DETR PyTorch Lightning Migration Plan


Borda:
### What maintainers gain

| Pain point today | After PTL migration |
|---|---|
| ~800 lines of boilerplate in `engine.py` + `main.py` (GradScaler, optimizer step, gradient clipping, AMP, DDP init, sampler logic) | All replaced by `Trainer` config — delete the code |
| Custom `defaultdict(list)` callback system in `RFDETR.train()` | PTL's typed, lifecycle-aware callback hooks |
| Manual multi-GPU setup (`init_distributed_mode`, custom samplers, `all_gather`, `reduce_dict`, `save_on_master`) | PTL handles entirely via `strategy="auto"` |
| Resume from checkpoint requires manual `start_epoch` tracking | `trainer.fit(..., ckpt_path=...)` — PTL resumes seamlessly |
| Two parallel config tiers: Pydantic (`config.py`) + argparse Namespace (`populate_args()`) | Pydantic only — `populate_args()` deleted |
| Testing training loop requires full run | PTL's `Trainer(fast_dev_run=1, limit_train_batches=...)` |
| `CocoEvaluator` wraps pycocotools internals; distributed sync is manual; `evalImgs` surgery for F1 sweep | Torchmetrics `MeanAveragePrecision` — distributed-aware, no internals access needed |
| `coco_eval.py` hand-rolls distributed gather + pycocotools patching | Deleted entirely |

### What users gain

| Capability | How |
|---|---|
| **Multi-GPU out of the box** | `rfdetr fit --trainer.devices 4` |
| **YAML experiments** | `rfdetr fit --config my_run.yaml` |
| **Resume training** | `rfdetr fit --ckpt_path output/last.ckpt` |
| **Logger integration** | `--trainer.logger WandbLogger` or YAML |
| **HP tuning** | Optuna, Ray Tune compatible |
| **Evaluate / predict** | `rfdetr validate/predict --ckpt_path best.ckpt` |
| **Versioned checkpoints** | PTL `ModelCheckpoint` auto-saves `last.ckpt` + `epoch=N.ckpt` |

---

## 🔒 Public API (backward-compatibility surface)

Everything below must remain importable and functionally equivalent after the migration, or have a `@deprecated` shim. This is what backward-compatible means in this context.

```
rfdetr/
  __init__.py
    ├─ RFDETRNano
    ├─ RFDETRSmall
    ├─ RFDETRMedium
    ├─ RFDETRBase
[...]

PR #757, opened 2026-02-27
RF-DETR PyTorch Lightning Migration 1/2

Borda:
Tracking progress on #709

This PR will cover only additions, and cleanup will be handled by another one...


Borda:
@copilot+gpt-5.4 investigate why the new ptl converge tests don't finish in given equal 10min timeout since in practice ptl was faster, check the configuration parity, could be that ptl testing is heavier? Start with compare model.train and model.train_ptl

PR #767, opened 2026-03-02
Feat/mlx inference

RacineAI-comp:
## What does this PR do?

  Adds native MLX segmentation inference so `RFDETRSegNano/Small/Medium/Large` work with `optimize_for_inference(backend="mlx")` on Apple Silicon.

  - New `MLXSegInferenceModel` with compiled FP16 forward pass (backbone → decoder with intermediates → seg head → masks)
  - New `SegHead` module (depthwise conv blocks + einsum mask generation)
  - Seg head weight conversion (`convert_seg_weights`) with key remapping
  - Decoder `return_intermediate` support to expose spatial features and per-layer hidden states
  - Automatic routing: `segmentation_head=True` in config → `MLXSegInferenceModel`, otherwise → `MLXInferenceModel`
  - Masks are resized to original image dimensions and returned as `sv.Detections(mask=...)`

  **Related Issue(s):** None

  ## Type of Change

  ## Testing

  **Test details:**

  - 14 new tests in `test_mlx_seg_inference.py`: weight conversion key remapping, conv transposition, num_blocks auto-detection, decoder `return_intermediate` shapes, `SegHead` output shapes, `build_seg_head` from dict, postprocess output
  keys/shapes/values
  - 2 new routing tests in `test_mlx_inference.py`: verifies seg config routes to `MLXSegInferenceModel` and det config routes to `MLXInferenceModel`
  - Full suite: 331 pass, 2 skipped, 0 failures (`pytest src/ tests/ -n 2 -m "not gpu"`)
  - `pre-commit run --all-files` clean
  - Verified on real image with `RFDETRSegLarge`: correct person/couch/remote detections with masks via MLX backend

  ## Checklist

  ## Additional Context

  Tested on Apple M4 Pro. The segmentation pipeline reuses the same backbone and decoder as the detection MLX backend, adding only the seg head and `return_intermediate` plumbing. No changes to existing detection inference behavior.


isaacrob:
how does the latency compare to existing implemenations such as https://github.com/Blaizzy/mlx-vlm/blob/5034c60932827a5baf39d46c4ef7ff70fa49d044/mlx_vlm/models/rfdetr/README.md?

Borda:
Hi @RacineAI-comp — apologies for letting this PR sit without a proper decision for so long, and thank you for the substantial work here (the seg head port and the `return_intermediate` plumbing in particular are clean).

We're going to close this rather than merge it, for reasons that are about scope, not code quality. To be clear up front: we'd be happy to enable MLX in RF-DETR — just not at the cost of a complete re-implementation of the model, and not without benchmarks demonstrating that it actually performs better than what we already ship.

- **Apple Silicon is already covered twice.** RF-DETR runs on the PyTorch MPS backend (eager and via `inference(compile=True)`), and exports to CoreML `.mlpackage` (`format="coreml"`, FP16 for ANE) as well as `format="executorch", backend="coreml"`. MLX would be a third runtime for the same hardware.
- **No benchmark justifies the third path.** @isaacrob asked in May how latency compares to existing implementations and that's still open. Without numbers we can't tell whether MLX beats MPS-compiled or CoreML on the same Mac.
- **~3.4k lines of hand-ported model code is a standing maintenance cost.** Every backbone, decoder, or head change would need a matching MLX port plus weight-key remapping in `convert_weights.py`; the `mlx`-marked tests only run on a macOS runner, so drift would surface late (patch coverage lands at 1.27% today).
- It also wires `backend="mlx"` and `_mlx_model` state directly into `RFDETR` instead of going through the exporter registry pattern the other formats use.

If you'd like to make the case, the thing that would change our mind is a benchmark table on one Mac and one model (e.g. `rfdetr-small` at 560): MPS eager, MPS `inference(compile=True)`, CoreML FP16, and MLX FP16 — median latency and peak memory. A clear MLX win there and we'd happily reopen and work out a lighter way to fit it into the package. In the meantime the port is well suited to a standalone package, or as a contribution to the existing `mlx-vlm` RF-DETR implementation, and we'll point people at it from #954.

Closing — please reopen with the benchmark if MLX shows a meaningful advantage.

Issue #406, opened 2025-10-10
RF-DETR incompatible with `torch.export.export`


olokobayusuf:
### Search before asking

- [x] I have searched the RF-DETR issues and found no similar bug report.

### Bug

`torch.export.export` is the recommended way to export models to other inference frameworks (CoreML, LiteRT, etc), but it breaks:

```py
from rfdetr import RFDETRBase
from torch import randn
from torch.export import export
from torch.nn import Module

# Create model
rf_detr = RFDETRBase()
model: Module = rf_detr.model.model.cpu()
model.export()
input_size: int = rf_detr.model.resolution

# Export
exported_program = export(
    model,
    args=(randn(1, 3, input_size, input_size),)
)
```

Exception:
```
.venv/lib/python3.12/site-packages/torch/export/__init__.py:319: in export
    raise e
.venv/lib/python3.12/site-packages/torch/export/__init__.py:286: in export
    return _export(
.venv/lib/python3.12/site-packages/torch/export/_trace.py:1164: in wrapper
    raise e
.venv/lib/python3.12/site-packages/torch/export/_trace.py:1130: in wrapper
    ep = fn(*args, **kwargs)
         ^^^^^^^^^^^^^^^^^^^
.venv/lib/python3.12/site-packages/torch/export/exported_program.py:123: in wrapper
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
.venv/lib/python3.12/site-packages/torch/export/_trace.py:2176: in _export
    ep = _export_for_training(
.venv/lib/python3.12/site-packages/torch/export/_trace.py:1164: in wrapper
    raise e
.venv/lib/python3.12/site-packages/torch/export/_trace.py:1130: in wrapper
    ep = fn(*args, **kwargs)
         ^^^^^^^^^^^^^^^^^^^
.venv/lib/python3.12/site-packages/torch/export/exported_program.py:123: in wrapper
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
.venv/lib/python3.12/site-packages/torch/export/_trace.py:2037: in _export_for_training
    export_artifact = export_func(
.venv/lib/python3.12/site-packages/torch/export/_trace.py:1979: in _non_strict_export
    aten_export_artifact = _to_aten_func(  # type: ignore[operator]
.venv/lib/python3.12/site-packages/torch/export/_trace.py:1770: in _export_to_aten_ir_make_fx
[...]

PR #812, opened 2026-03-17
fix: use static normalized_shape in projector LayerNorm for ONNX export compatibility

vittorio-prodomo:
## Summary

Replace `(x.size(3),)` with `self.normalized_shape` in the projector's
`LayerNorm.forward()`. This is functionally identical — `self.normalized_shape`
is set to `(channel_dim,)` in `__init__`, which always equals `x.size(3)` at
runtime — but makes the ONNX export succeed on PyTorch 2.9+.

## Problem

Starting with PyTorch 2.9, `torch.onnx.export` defaults to the new
dynamo-based exporter (`dynamo=True`). This exporter treats `x.size(3)` as
a symbolic value that could vary across inputs, and refuses to use it where
a static shape is required (the `normalized_shape` argument to
`F.layer_norm`).

Falling back to the legacy TorchScript exporter (`dynamo=False`) also fails
on PyTorch 2.9 because the tracer is now stricter about `prim::ListConstruct`
nodes with non-constant inputs.

On PyTorch < 2.9, the legacy tracer evaluated `x.size(3)` eagerly during
tracing and embedded the result as a constant, so the export succeeded.
This is most likely why the issue was not previously reported.

The error:

```
torch.onnx.errors.SymbolicValueError: Failed to export a node ... because
it is not constant. Please try to make things (e.g. kernel sizes) static
if possible.
```

at `rfdetr/models/backbone/projector.py:43`.

## Fix

```diff
- x = F.layer_norm(x, (x.size(3),), self.weight, self.bias, self.eps)
+ x = F.layer_norm(x, self.normalized_shape, self.weight, self.bias, self.eps)
```

`self.normalized_shape` is a tuple `(channel_dim,)` assigned in `__init__`
and never changes. Both the dynamo and legacy exporters see it as a constant.

## Impact

- Fixes ONNX export for all RF-DETR variants (detection and segmentation,
  all sizes) on PyTorch 2.9+
- No behavioral change at inference time
- Related: #406 (this removes one of several obstacles to `torch.export` compat)

### DIFF
--- a/src/rfdetr/models/backbone/projector.py
+++ b/src/rfdetr/models/backbone/projector.py
@@ -40,7 +40,7 @@ class LayerNorm(nn.Module):
         TODO: this is a hack to avoid overflow when using fp16
         """
         x = x.permute(0, 2, 3, 1)
-        x = F.layer_norm(x, (x.size(3),), self.weight, self.bias, self.eps)
+        x = F.layer_norm(x, self.normalized_shape, self.weight, self.bias, self.eps)
         x = x.permute(0, 3, 1, 2)
         return x
 
--- a/tests/models/backbone/test_projector.py
+++ b/tests/models/backbone/test_projector.py
@@ -26,48 +26,47 @@ import torch
 from rfdetr.models.backbone.projector import LayerNorm
 
 
-def test_layernorm_forward_preserves_shape():
-    """LayerNorm should preserve input shape (B, C, H, W)."""
-    channels = 64
-    ln = LayerNorm(channels)
-    x = torch.randn(2, channels, 8, 8)
-    out = ln(x)
-    assert out.shape == x.shape
+class TestProjectorLayerNorm:
+    def test_layernorm_forward_preserves_shape(self) -> None:
+        """LayerNorm should preserve input shape (B, C, H, W)."""
+        channels = 64
+        ln = LayerNorm(channels)
+        x = torch.randn(2, channels, 8, 8)
+        out = ln(x)
+        assert out.shape == x.shape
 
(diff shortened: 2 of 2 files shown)


vittorio-prodomo:
Apologies for the noise — I accidentally included an unrelated change (centralized weight cache directory) in this PR alongside the ONNX export fix. I've rebased to remove it, so this PR now contains only the single-line `normalized_shape` fix in `projector.py` as intended.

src/rfdetr/models/backbone/projector.py
```diff
@@ -40,7 +40,7 @@ def forward(self, x):
         TODO: this is a hack to avoid overflow when using fp16
         """
         x = x.permute(0, 2, 3, 1)
-        x = F.layer_norm(x, (x.size(3),), self.weight, self.bias, self.eps)
+        x = F.layer_norm(x, self.normalized_shape, self.weight, self.bias, self.eps)
```
Borda, line 43: @vittorio-prodomo can we pls have test to support this fix

vittorio-prodomo: absolutely. I've added regression tests in `tests/models/backbone/test_projector.py`:

- **`test_layernorm_forward_preserves_shape`** — sanity check that the forward pass preserves `(B, C, H, W)` shape
- **`test_layernorm_normalized_shape_is_static`** — asserts `normalized_shape` is a static tuple, not derived from a tensor at runtime
- **`test_layernorm_onnx_export`** — end-to-end ONNX export with dynamic axes, confirming no `SymbolicValueError` on PyTorch 2.9+

PR #814, opened 2026-03-18
feat(training): add batch_size='auto' with train-step probe 

HaochengLu:
## What does this PR do?

This PR adds support for **`batch_size="auto"`** in `model.train()`. Before training starts, a lightweight memory probe runs on the target CUDA device to:

- **Find a safe micro-batch size** using an exponential search + binary search (train-step forward + loss + backward only; no dataset load).
- **Recommend `grad_accum_steps`** so that `effective_batch_size = micro_batch_size × grad_accum_steps` meets a configurable target (default 16 via `auto_batch_target_effective`).
- **Resolve and log** the final `batch_size` and `grad_accum_steps` so the rest of the training pipeline runs unchanged with integer config.

Users can call:

```python
model.train(
    dataset_dir="...",
    output_dir="...",
    batch_size="auto",
    auto_batch_target_effective=16,  # optional
)
```

and get a clear recommendation before the first epoch (e.g. *"safe micro-batch = 3, grad_accum_steps = 4, effective_batch_size = 12"*). The probe is **train-step only**; validation/test (especially segmentation mask eval) may use more memory and are out of scope for this PR.

**Related Issue(s):** Relates to #709

---

## Type of Change

- **New feature** (non-breaking change that adds functionality)

---

## Testing

**Test details:**

- **Config:** `tests/models/test_config.py` — `batch_size="auto"` and `auto_batch_target_effective` are accepted; `batch_size` / `grad_accum_steps` / `auto_batch_target_effective` reject non-positive values.
- **Train entry:** `tests/training/test_detr_shim.py` — when `batch_size="auto"`, `resolve_auto_batch_config` is called with the correct model context and train config; resolved `batch_size` and `grad_accum_steps` are passed to `RFDETRModule` and `RFDETRDataModule`; existing tests (e.g. resume, ckpt_path) still pass with a mock config that has integer `batch_size`.
- **Auto-batch module:** `tests/training/test_auto_batch.py` — `recommend_grad_accum_steps` rounding, `probe_max_micro_batch` behavior (exponential + binary search, failure at micro_batch=1 raises with clear message), and `resolve_auto_batch_config` requiring CUDA and returning expected `AutoBatchResult` fields.

Run the relevant tests:

```bash
python -m pytest tests/models/test_config.py tests/training/test_detr_shim.py tests/training/test_auto_batch.py -v
```

---

## Checklist

*(Documentation: `TrainConfig` now documents `batch_size: int | "auto"` and `auto_batch_target_effective`; log messages state that the probe is train-step only and that val/test may require more memory.)*

---

### Why

RF-DETR currently requires manual trial-and-error for `batch_size` and `grad_accum_steps` across different GPUs and resolutions. This PR adds a first step toward a more ergonomic, YOLO-like training experience: one-shot pre-training probe, then normal training with resolved config.

### Design (minimal surface)

- **Single insertion point:** In `RFDETR.train()`, after `get_train_config(**kwargs)`, if `config.batch_size == "auto"`, call `resolve_auto_batch_config(...)` and overwrite `config.batch_size` and `config.grad_accum_steps` before building `RFDETRModule` / `RFDETRDataModule` / `build_trainer`. No changes to the training loop, datamodule, or trainer logic.
- **New module:** `src/rfdetr/training/auto_batch.py` — `AutoBatchResult`, `probe_max_micro_batch`, `recommend_grad_accum_steps`, `resolve_auto_batch_config`. CUDA-only in v1; no multi-GPU or eval probing.
- **Config:** `TrainConfig` extended with `batch_size: int | Literal["auto"] = 4` and `auto_batch_target_effective: int = 16`, plus validators so that integer `batch_size` and both accumulation fields are ≥ 1.

### Scope (v1)

- **In scope:** `batch_size="auto"`, CUDA probe, train-step-safe micro-batch and recommended grad_accum_steps, detection and segmentation (same mechanism; segmentation may get a smaller safe batch).
- **Out of scope:** Auto resolution tuning, gradient checkpointing toggling, eval/test memory probing, mid-training OOM recovery, multi-device probing.

### Example log output

```
[auto-batch] device=NVIDIA GeForce RTX 4090 segmentation=False resolution=672
[auto-batch] safe_micro_batch=3 grad_accum_steps=4 effective_batch_size=12
[auto-batch] This probe estimates train-step-safe micro-batch only.
[auto-batch] Validation/test (especially segmentation mask eval) may require more memory.
[auto-batch] resolved train config: batch_size=3 grad_accum_steps=4 effective_batch_size=12
```

Relates to #709.


src/rfdetr/training/auto_batch.py
```diff
@@ -0,0 +1,224 @@
...
+            {
+                "boxes": torch.tensor([[0.5, 0.5, 0.2, 0.2]], dtype=torch.float32, device=device),
+                "labels": torch.tensor([min(1, max_label)], dtype=torch.int64, device=device),
+                "image_id": torch.tensor(idx, dtype=torch.int64, device=device),
+                "orig_size": torch.tensor([resolution, resolution], dtype=torch.int64, device=device),
+                "size": torch.tensor([resolution, resolution], dtype=torch.int64, device=device),
+                "iscrowd": torch.tensor([0], dtype=torch.int64, device=device),
+                "area": torch.tensor([0.04], dtype=torch.float32, device=device),
+            }
+        )
```
Borda: `_make_synthetic_batch` doesn't add a `"masks"` key to targets. When `segmentation_head=True`, the criterion's `loss_masks` calls `t["masks"][j]` after matching — you'll get a `KeyError` for every segmentation model. Add `segmentation_head: bool = False` to this function and include `"masks": torch.zeros(1, resolution, resolution, dtype=torch.bool, device=device)` per target when it's set.

src/rfdetr/training/auto_batch.py
```diff
@@ -0,0 +1,224 @@
...
+    return samples, targets
+
+
+def _probe_step(
```
Borda, line 102: `_probe_step` doesn't pass a `segmentation_head` flag through to `_make_synthetic_batch`. Once you add the flag to `_make_synthetic_batch`, thread it here so the synthetic targets match what the criterion expects.

src/rfdetr/training/auto_batch.py
```diff
@@ -0,0 +1,224 @@
...
+        criterion.zero_grad(set_to_none=True)
+        return True
+    except RuntimeError as exc:
+        if _is_cuda_oom(exc):
+            torch.cuda.empty_cache()
+            return False
+        raise
+
+
+def probe_max_micro_batch(
```
Borda, line 146: `probe_max_micro_batch` has no docstring — describe the exponential + binary search algorithm, what `safety_margin` controls, and that `RuntimeError` is raised when `micro_batch_size=1` fails.

src/rfdetr/training/auto_batch.py
```diff
@@ -0,0 +1,224 @@
...
+        while lo <= hi:
+            mid = (lo + hi) // 2
+            if _probe_step(model, criterion, mid, resolution, device, num_classes, amp):
+                lower_ok = mid
+                lo = mid + 1
+            else:
+                hi = mid - 1
+
+        safe_micro_batch = max(1, math.floor(lower_ok * safety_margin))
+        return min(safe_micro_batch, lower_ok)
```
Borda: `min(safe_micro_batch, lower_ok)` is always `safe_micro_batch` since `math.floor(lower_ok * safety_margin) ≤ lower_ok` for `safety_margin ≤ 1.0`. The `min()` can be removed; a brief comment on the invariant is enough.

src/rfdetr/training/auto_batch.py
```diff
@@ -0,0 +1,224 @@
...
+# RF-DETR
+# Copyright (c) 2025 Roboflow. All Rights Reserved.
+# Licensed under the Apache License, Version 2.0 [see LICENSE for details]
+# ------------------------------------------------------------------------
+
+from __future__ import annotations
+
+import math
+from dataclasses import dataclass
+from typing import Any, Dict, List, Tuple
```
Borda: `from typing import Any, Dict, List, Tuple` — `Dict`/`List`/`Tuple` are legacy; with `from __future__ import annotations` already present and Python ≥3.10, use `dict`, `list`, `tuple` directly. `Any` is still needed.

src/rfdetr/training/auto_batch.py
```diff
@@ -0,0 +1,224 @@
...
+
+def recommend_grad_accum_steps(safe_micro_batch: int, target_effective_batch: int) -> int:
+    if safe_micro_batch < 1:
+        raise ValueError("safe_micro_batch must be >= 1.")
+    if target_effective_batch < 1:
+        raise ValueError("target_effective_batch must be >= 1.")
+    return max(1, math.ceil(target_effective_batch / safe_micro_batch))
+
+
+def resolve_auto_batch_config(
```
Borda, line 279: `resolve_auto_batch_config` has no docstring. Document `model_context` (what type and attributes are expected), `safety_margin`, `max_micro_batch`, the return type, and the `RuntimeError` raised on non-CUDA.

Borda (commented):
@HaochengLu the two-phase exponential + binary search probe is a clean approach, and the single insertion point in `RFDETR.train()` — resolve, overwrite, then hand off to the PTL stack — is exactly right. The scope is well-contained and the detection path works.

Two things need sorting before I merge.

The probe crashes with a `KeyError: 'masks'` for any segmentation model. When `model_config.segmentation_head=True`, `build_criterion_and_postprocessors` adds `"masks"` to the losses list (`lwdetr.py:443`), and `SetCriterion.loss_masks` accesses `t["masks"][j]` after the Hungarian matcher produces non-empty indices for your synthetic one-box-per-image target (`criterion.py:403`). The test at line 87 passes `segmentation_head=True` but mocks out `probe_max_micro_batch` entirely, so the real criterion path is never exercised. Fix: add a `segmentation_head: bool = False` parameter to `_make_synthetic_batch` and include a `"masks": torch.zeros(1, resolution, resolution, dtype=torch.bool, device=device)` entry per target when it's `True`. Thread the flag through `_probe_step` → `probe_max_micro_batch` → `resolve_auto_batch_config` (it's already available via `model_config.segmentation_head`). Then add a test that runs `_probe_step` with a real segmentation criterion instead of mocking it, so this path stays covered.

Minor: `probe_max_micro_batch`, `recommend_grad_accum_steps`, and `resolve_auto_batch_config` are missing Google-style docstrings — these are required for all public functions per the project guidelines. `AutoBatchResult` should have a class-level docstring too. Also, `auto_batch.py` imports `Dict`, `List`, `Tuple` from `typing` — the project is Python ≥3.10 and the rest of the codebase uses built-in generics (`dict`, `list`, `tuple`); the `from __future__ import annotations` is already there, so the legacy imports can go. One small redundancy: `min(safe_micro_batch, lower_ok)` on line 158 is always equal to `safe_micro_batch` because `safety_margin ≤ 1.0` guarantees `safe_micro_batch ≤ lower_ok` — the `min()` can be dropped.

src/rfdetr/training/auto_batch.py
HaochengLu: Thanks for the detailed review and for confirming the approach. I’ve fixed all of the issues you raised, and I’ll submit a new PR shortly.

HaochengLu:
@Borda Thanks for the review. I've applied the fixes and updated the PR.

isaacrob:
Memory use varies significantly with number of targets and random selected resolution. In fact, memory is unbounded as targets increase. I would recommend either using a parameter specifying max number of targets or referencing the actual dataset for that number. I would also ensure the step uses the max resolution used in multi scale augmentation.

isaacrob:
Additionally, EMA takes up some memory.

HaochengLu:
> Memory use varies significantly with number of targets and random selected resolution. In fact, memory is unbounded as targets increase. I would recommend either using a parameter specifying max number of targets or referencing the actual dataset for that number. I would also ensure the step uses the max resolution used in multi scale augmentation.

Thanks for the feedback. I’ve addressed both points:

Max number of targets
I added a config parameter auto_batch_max_targets_per_image (default: 100). The probe now builds synthetic batches with that many targets per image so that matcher and loss memory reflect a worst case (or a user-defined upper bound). So we no longer assume 1 target per image; users can set this to match their dataset or leave the default.

Max resolution for multi-scale
When multi_scale=True (and we’re not using do_random_resize_via_padding), the probe now uses the maximum of the multi-scale augmentation scales (same compute_multi_scale_scales as in training) as the probe resolution, instead of the base model_config.resolution. So the probe step uses the same max resolution that can appear during training.

I also added an EMA headroom (auto_batch_ema_headroom, default 0.7) when use_ema=True, since the EMA copy uses extra memory.

Borda:
> Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](https://cla-assistant.io/roboflow/rf-detr?pullRequest=814) before we can accept your contribution.
> **2** out of **3** committers have signed the CLA.

@HaochengLu ^^ 🙏

HaochengLu:
> > Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](https://cla-assistant.io/roboflow/rf-detr?pullRequest=814) before we can accept your contribution.
> > **2** out of **3** committers have signed the CLA.
> 
> @HaochengLu ^^ 🙏

Just signed🌹

Borda:
> Just signed🌹

seems there is some delay, or can you pls share the screenshot here?

HaochengLu:
> > Just signed🌹
> 
> seems there is some delay, or can you pls share the screenshot here?

image

Borda:
> > > Just signed🌹
> > 
> > 
> > seems there is some delay, or can you pls share the screenshot here?
> 
> image

thank you!

Issue #376, opened 2025-10-01
Official ONNX Export Script for Dynamic Batching


Mahaaveer:
### Search before asking

- [x] I have searched the RF-DETR [issues](https://github.com/roboflow/rf-detr/issues) and found no similar feature requests.

### Description

Dear RF-DETR Team,

First off, thank you for creating and maintaining this fantastic project. The RF-DETR Nano model is incredibly efficient and i am looking to use it in a multi-stream video analytics pipeline.

**The Goal:**

The goal is to convert the RF-DETR Nano model to a TensorRT engine that supports dynamic batch sizes. This is critical for maximizing GPU throughput in a real-time application that processes frames from multiple camera streams simultaneously.

**The Problem:**

The current default export method (model.export()) produces a static ONNX file with a hardcoded batch size of 1. When i tried to convert this to a TensorRT engine with dynamic shapes, trtexec fails as expected.

I spent considerable time trying to create a custom export script to produce a valid, dynamic ONNX file, but have run into a series of issues that suggest the model's architecture contains patterns that are incompatible with the standard torch.onnx.export tracing process when dynamic axes are used.

**Summary of Attempts and Findings:**

**Initial Attempt (Dynamic Axes)**: I created a custom export script using torch.onnx.export with dynamic_axes set for the batch dimension. This failed during trtexec conversion with a reshape would change volume error, indicating an internal Reshape operation was using a hardcoded shape based on the trace-time batch size of 1.

**Input Type Mismatch**: We then discovered the model's forward pass expects a custom NestedTensor object, not a raw tensor. The tracer was failing because it was providing the wrong input type.

**Wrapper Approach**: I implemented the standard "ONNX Export Wrapper" pattern. This correctly constructs the NestedTensor from raw tensor inputs, satisfying the model's API. This successfully resolved the input type errors.

**Unsupported Operator**: The wrapper allowed the tracer to go deeper, where it found an unsupported operator aten::_upsample_bicubic2d_aa. I solved this by registering a custom symbolic function to map it to the standard ONNX Resize operator.

**Final trtexec Error**: Even with a seemingly perfect ONNX file, trtexec ultimately fails during parsing with an INVALID_NODE error on a ConvTranspose layer, stating that the number of input channels cannot be dynamic. We have also seen Concat layer errors where a cls_token with a static batch size of 1 is concatenated with feature maps that have a dynamic batch size.

**Conclusion:**

[...]

PR #79, opened 2025-03-29
[ONNX] Add dynamic shapes support (& in-browser inference w/ Transformers.js)

xenova:
# Description

This PR adds support for exporting RF-DETR to ONNX with dynamic input and output shapes. This means you can now supply images of variable batch_size, height, and width, both for backbone-only and full exports (provided width and height are divisible by 56).

I have uploaded the checkpoints (along with various quantizations) to the Hugging Face Hub:
- Base: https://huggingface.co/onnx-community/rfdetr_base-ONNX
- Large: https://huggingface.co/onnx-community/rfdetr_large-ONNX

## Type of change

-   [x] New feature (non-breaking change which adds functionality)

### Exporting:
```py
from rfdetr import RFDETRBase

model = RFDETRBase()

model.export()
model.export(backbone_only=True)
```

Similarly for `RFDETRLarge`.

### Testing:

Running with Transformers.js, the model outputs the correct response (i.e., valid model, and valid dynamic shape support). See PR at https://github.com/huggingface/transformers.js/pull/1260.

## Any specific deployment considerations

I would recommend further optimizing with the amazing [onnxslim](https://github.com/inisis/OnnxSlim) library; you can reduce model size quite a bit (eliminating tied weights and redundant ops).

Base:

```
+--------------------+-----------------------------------------+--------------------------------------+
|     Model Name     |          inference_model.onnx           |       ./final-base/model.onnx        |
+--------------------+-----------------------------------------+--------------------------------------+
|     Model Info     |       Op Set: 17 / IR Version: 8        |      Op Set: 17 / IR Version: 8      |
+--------------------+-----------------------------------------+--------------------------------------+
|     IN: input      |  float32: ('batch_size', 3, 'height',   | float32: ('batch_size', 3, 'height', |
|                    |                'width')                 |               'width')               |
|     OUT: dets      |      float32: ('Concatdets_dim_0',      |   float32: ('batch_size', 300, 4)    |
|                    | 'Concatdets_dim_1', 'Concatdets_dim_2') |                                      |
|    OUT: labels     |      float32: ('Addlabels_dim_0',       |   float32: ('batch_size', 300, 91)   |
|                    |         'Addlabels_dim_1', 91)          |                                      |
+--------------------+-----------------------------------------+--------------------------------------+
|        Add         |                   186                   |                 181                  |
|        And         |                    1                    |                  1                   |
|        Cast        |                   109                   |                  5                   |
|       Concat       |                   155                   |                  88                  |
[...]

PR #871, opened 2026-03-25
Fix ONNX export for dynamic batch dimensions

svengoluza:
## What does this PR do?

ONNX export silently baked spatial dimensions and batch size into the graph as fixed constants. When users tried to run the exported model with a different batch size at inference time, it would fail because the ONNX graph contained hardcoded values derived from the export-time batch size.

Three root causes:

1. **`LayerNorm.forward` used `x.size(3)` instead of `self.normalized_shape`** — The ONNX tracer captured the concrete integer from the export-time tensor shape and embedded it as a constant node. Switching to `self.normalized_shape` (already stored on the module) gives the tracer a static attribute it can reference symbolically.

2. **`spatial_shapes` was built as a Python list, then converted with `torch.as_tensor()`** — The tracer never saw the `h, w` values flow through tensor operations, so it treated them as constants. Building `spatial_shapes` directly as a tensor with index assignment (`spatial_shapes[lvl, 0] = h`) lets the tracer track the symbolic relationship between the input spatial dims and downstream uses.

3. **`gen_encoder_output_proposals` created `valid_H`/`valid_W` via Python list comprehensions** — `torch.tensor([H_ for _ in range(N_)])` bakes the batch-dependent value `N_` as a fixed list length. Replacing this with `H_.expand(N_)` keeps the batch dimension symbolic in the graph.

Additionally, a `dynamic_batch` parameter is added to `RFDETR.export_onnx()` and the standalone export CLI. When enabled, it marks the batch axis (`dim 0`) as dynamic on all input and output names, allowing the exported model to accept variable batch sizes at runtime.

**Related Issue(s):** https://github.com/roboflow/rf-detr/issues/376 , https://github.com/roboflow/rf-detr/pull/79

## Type of Change

## Testing


src/rfdetr/models/transformer.py
```diff
@@ -224,12 +224,22 @@ def forward(self, srcs, masks, pos_embeds, refpoint_embed, query_feat):
         src_flatten = []
         mask_flatten = [] if masks is not None else None
         lvl_pos_embed_flatten = []
-        spatial_shapes = []
+        # Build spatial_shapes as a tensor directly so that the ONNX tracer
+        # can track h/w symbolically instead of baking them in as constants.
+        spatial_shapes = torch.empty(
+            (len(srcs), 2),
+            device=srcs[0].device,
+            dtype=torch.long,
```
Borda:
```suggestion
            dtype=torch.long
```

Borda (approved):

PR #891, opened 2026-03-30
fix: correct Image.open URL usage in README examples

farukalamai:
## What does this PR do?

Fix broken code examples in README.md. `PIL.Image.open()` does not accept URLs directly — it raises `FileNotFoundError`. Updated all 4 examples to use `requests.get()` to fetch the image first, matching the docs website.

**Related Issue(s):** None (bug found while testing examples)

## Type of Change

## Testing

**Test details:**
Confirmed `Image.open("https://...")` raises `FileNotFoundError`. Confirmed the fix works correctly.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,7 @@ from rfdetr.assets.coco_classes import COCO_CLASSES
 
 model = RFDETRMedium()
 
-image = Image.open("https://media.roboflow.com/dog.jpg")
+image = Image.open(requests.get("https://media.roboflow.com/dog.jpg", stream=True).raw)
 detections = model.predict(image, threshold=0.5)
 
 labels = [f"{COCO_CLASSES[class_id]}" for class_id in detections.class_id]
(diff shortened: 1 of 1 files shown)


Borda (approved):
Well, this is correct, and I will merge it. However, the better way would be demo passing the string directly to the predict, which shall also be supported... maybe suggestion for follow-up PR :)

farukalamai:
Thanks @Borda for your feedback! I'll work on a follow-up PR that simplifies the examples to pass the string directly to `predict()`.                                                      
                                                                                                                                                                                     
One thing, we still need the image object for `sv.BoxAnnotator().annotate(image, detections)`. I was thinking about storing the original image on the returned detections (via detections.data["source_image"]) so users can access it without loading the image separately. That way, we could drop both **from PIL import Image** and  **import requests** from the examples entirely. 

What do you think? Should I open an issue first?

Borda:
That sounds good, if you want to take it, just open a new PR and link this comment...

farukalamai:
Hi @Borda, opened the follow-up PR: #892

Issue #173, opened 2025-04-19
Cannot Export to Tesnorflow or Tflite to run on mobile


apcarita:
ONNX-TF only supports up to opset 13 for some operators (like Where), but
Your model uses operators (like aten::scaled_dot_product_attention) that are only supported in ONNX opset 14 or higher.
Summary of your situation:

If you export with opset 13: PyTorch cannot export your model because it uses newer operators.
If you export with opset 16: ONNX-TF cannot convert your model because it does not support some ops at that version.

dsbyprateekg:
@apcarita have you tried this PR for tflite conversion: [https://github.com/roboflow/rf-detr/pull/45](url)

gpokat:
seems to work for me, will test on device and update.
thanks !

dsbyprateekg:
> seems to work for me, will test on device and update. thanks !

how is the performance of the tflite model in mobile device?

apcarita:
Awesome thank you!

stoff99:
@gpokat do you have any update for us, how it performs on Android with tflite format? Would be very interesting. 
Thx

gpokat:
Guys, im also very curious, but the latest tests shows that we are lost accuracy during double converting from pytorch to tflite via onnx especially for quantized variant. Tested on qnx and performance is not much bigger then yolov5 with fp16 in compare.
Not tested on Android, but here a plenty of work to do before real benchmarking which i have no time for now.

herjanice:
Hi guys, Im also very interested in this, do anyone have an update on this?

matiasso:
Also curious on updates if anyone has experimented with Android/iOS, how is the performance on mobile devices?

isaacrob:
Someone on our team got nano to hit 120 fps on iPhone 16

matiasso:
@isaacrob Could you share some more details? Was it exported to ONNX or CoreML format? I suppose it was utilizing iPhone 16 GPU or NPU, because it's hard to believe those numbers are on CPU alone? 

I'm curious to see how it generalizes across different modern and older mobile phones, as (specifically in the Android world of mobile devices) there are thousands of varying hardware combinations out there, which will most likely still cause some devices to fallback to CPU inference, if the device does not have fancy modern NPUs/GPUs, which would support the exported Opset for example...

[...]

PR #45, opened 2025-03-25
code for exporting onnx model to tflite

dsbyprateekg:
Using this code, I am able to convert my fine tuned model from onnx to tflite format. I have also added the inference code. This code has been tested in google colab notebook with A100 GPU.

# Description

First I converted my fine tuned model to onnx format using below code-
from rfdetr import RFDETRBase
model = RFDETRBase(pretrain_weights="rf-detr_lego_model_30epochs.pth")
model.export()

Then I use following library for converting onnx model to tflite format: https://github.com/PINTO0309/onnx2tf
I also added inference code with converted tflite model to verify if its working as expected or not.

List any dependencies that are required for this change.

## code has been tested in google colab notebook with A100 GPU.

Please delete options that are not relevant.
-   [ ] This change requires a documentation update

## How has this change been tested, please provide a testcase or example of how you tested the change?
I have tested the inference using converted tflite model and its output is as same as original fine tuned model.

Issue #870, opened 2026-03-25
ModelConfig should accept `torch.device`


vmiller987:
### Search before asking

- [x] I have searched the RF-DETR issues and found no similar bug report.

### Bug

RF-DETR's `ModelConfig.device` field is typed as `Literal["cpu", "cuda", "mps"]` (`config.py:71`), which means it cannot accept `torch.device` objects (e.g., `torch.device("cuda:1")`) or device strings with a GPU index (e.g., "cuda:0", "cuda:1").

This makes it impossible to target a specific GPU through the RF-DETR API. The only workaround is setting the `CUDA_VISIBLE_DEVICES` environment variable, but this is a process global and has significant limitations.
- No concurrent multi-GPU inference in a single process. Two RF-DETR models cannot run on different GPUs simultaneously, since `CUDA_VISIBLE_DEVICES` affects the entire process.
- Device index remapping: Setting `CUDA_VISIBLE_DEVICES=2` causes physical GPU 2 to appear as `cuda:0`, which can conflict with other libraries or models in the same process that expect the original device mapping.
- Not composable: Any framework wrapping RF-DETR (e.g., an orchestration layer managing multiple backends on different GPUs) cannot use PyTorch's standard `torch.device("cuda:N")` pattern and must instead manipulate environment variables.

PyTorch's convention is to accept `torch.device` objects or strings like `cuda:0` for GPU selection. RF-DETR's training path already acknowledge this gap: `detr.py:305-320` absorbs and discards the device kwarg with a DeprecationWarning, telling users to use `CUDA_VISIBLE_DEVICES` instead. 

### Environment

RF-DETR: 1.6.0
OS: RHEL 9.7
Python: 3.13
PyTorch: 2.8 - 2.10
CUDA: 12.9
GPU: RTX 4090, 5090, Pro 6000

### Minimal Reproducible Example

```python
import torch
from rfdetr import RFDETRNano

# Fails, Pydantic rejects "cuda:1"
model = RFDETRNano(device="cuda:1")

# Fails, Pydantic rejects torch.device
model = RFDETRNano(device=torch.device("cuda:1"))

# Current workaround
import os

os.environ["CUDA_VISIBLE_DEVICES"] = "1"
model = RFDETRNano(device="cuda")
```

### Additional

- Change `ModelConfig.device` to accept `str | torch.device` (or at minimum `str` without the `Literal` restriction). 
[...]

PR #920, opened 2026-04-03
feat(export): add TFLite export support with FP32/FP16/INT8

mfazrinizar:
## What does this PR do?

Adds an ONNX → TFLite conversion pipeline for RF-DETR detection and segmentation models via `onnx2tf`.

Users can now export any RF-DETR model to TFLite format with a single call:

```python
from rfdetr import RFDETRNano

model = RFDETRNano()

# FP32 (recommended — also produces FP16 automatically)
model.export(format="tflite")

# INT8 with calibration images
model.export(format="tflite", quantization="int8", calibration_data="path/to/images/")
```

### Key changes

**New module: `src/rfdetr/export/_tflite/`**
- `converter.py`: Core `export_tflite()` function — converts ONNX to TFLite via onnx2tf Python API
- Supports `quantization="int8"` with calibration from an image directory, `.npy` file, or NumPy array
- `max_images` parameter controls calibration sample count (default: 100)
- Applies `output_signaturedefs=True` fix for segmentation models (onnx2tf node naming issue with leading `/` characters violating TF saved_model naming pattern)
- Monkey-patches `onnx2tf`'s `download_test_image_data()` to use local calibration data instead of downloading from GitHub (which fails in CI, firewalls, air-gapped systems)
- NumPy `allow_pickle` compatibility shim for onnx2tf 1.x releases

**Modified files**
- `src/rfdetr/detr.py`: Added `format="tflite"` routing in `export()` with `quantization`, `calibration_data`, `max_images`, `output_dir` parameters
- `src/rfdetr/export/main.py` / `__init__.py`: Wired up `export_tflite` import
- `pyproject.toml`: Added `onnx2tf>=1.26.0` and `tf-keras>=2.16.0` to `[onnx]` optional dependencies
- `docs/learn/export.md`: TFLite usage guide, calibration explanation, parameter reference, FP32/FP16 always-produced behavior note

**Related Issue(s):**  Related to #173 and this [PR](https://github.com/roboflow/rf-detr/pull/45)

## Type of Change

## Testing

**Test details:**

56 unit tests in `tests/export/test_tflite_export.py` (all CPU, no GPU required):

- **Format routing**: `export(format="tflite")` correctly dispatches to `export_tflite`
- **Quantization options**: Validates `None`, `"fp32"`, `"fp16"`, `"int8"` modes; rejects invalid values
- **Calibration data preparation**: Tests random noise generation, `.npy` file loading, NumPy array input, and directory-based image loading with resize
- **`max_images` parameter forwarding**: Verifies `max_images` flows through `detr.py` → `export_tflite()` → `_prepare_calibration_data()` → `_load_calibration_images()`
- **Error handling**: Missing ONNX file, missing calibration directory, empty image directory
- **INT8 warning**: Emits warning when no calibration data provided for INT8

## Checklist


mfazrinizar:
Hey @Borda ,

While I'm looking at the CI failures and Copilot feedback for this onnx2tf implementation, I had a thought about the overall TFLite export strategy.

What are your thoughts on potentially pivoting and using Google's [litert-torch](https://www.google.com/search?q=%5Bhttps://github.com/google-ai-edge/litert-torch%5D(https://github.com/google-ai-edge/litert-torch)) (LiteRT for PyTorch) instead of routing through ONNX as a middleman?

I'm still just in the planning phase for it, but exporting directly to TFLite might be a cleaner pipeline in the long run and would help us avoid some of this heavy dependency overhead (like the tf-keras and onnx2tf versioning problems we're seeing here).

I plan to open a fresh PR for it and we can close this one out. Let me know what you think!

farukalamai (changes requested):
Hi @Borda, I reviewed this PR and wanted to mention a **few issues** I noticed:

- `_patch_validation_download` in `converter.py` patches `onnx2tf` sub-modules only if they're already in `sys.modules`. In a fresh process they won't be, so the patch silently does nothing. The sub-modules need to be force-imported before the patching happens.
- `rfdetr[tflite]` doesn't include `tensorflow` itself — `tf-keras` won't pull it in. Someone doing a clean install will hit a confusing runtime failure with no clear message pointing to the missing package.
- The CLI path in `export/main.py` looks broken — `--tflite`, `--quantization`, `--calibration-data`, and `--max-images` are never actually registered in the argument parser, so CLI users have no way to trigger TFLite export right now.
 - `onnx2tf>=1.25.0,<3.0.0` is too wide — the monkey-patches target 1.x internals specifically, and 2.x has different output naming and a different conversion path.

---

**On the `litert-torch` idea — I think @mfazrinizar is right.**

Dragging in TensorFlow for this is a lot of weight, and the monkey-patching is a direct consequence of routing through `onnx2tf`. `litert-torch` cuts all of that out and keeps the pipeline much simpler. Since the author is already planning a fresh PR for it, closing this one makes sense.

Borda (approved):

Borda:
Hi @mfazrinizar — your instinct to look at `litert-torch` was exactly right, and I appreciate you raising it. I spent time researching whether we should pivot before deciding to keep this PR moving. Here's what I found, so the decision is transparent.

`ai-edge-torch` (which you may have seen referenced) is now deprecated. Its successor, `litert-torch` v0.8.0, reached Beta in January 2026 and represents the clean future path: `torch.export()` → Core ATen ops → JAX/XLA tracing → TFLite flatbuffer directly, with no ONNX, no TensorFlow, and none of the monkey-patching this PR currently needs. If it worked today, we'd drop roughly five heavy dependencies (`onnx`, `onnxruntime`, `onnx2tf`, `tensorflow`, `tf-keras`) and the conversion pipeline would be far simpler.

The blocker is `litert-torch` issue [#870](https://github.com/google-ai-edge/LiteRT/issues/870) — currently open. The XLA tracer raises `jax.errors.TracerBoolConversionError` whenever `aten.slice` receives a symbolic `SymInt` dimension. Static shapes work reliably; dynamic shapes do not. RF-DETR's variable batch size and multi-scale positional encodings land squarely in the broken path — the backbone and encoder would fail during tracing. (The 300-query fixed output head would be fine, but that's not enough on its own.)

So: your directional read was correct, but `litert-torch` isn't ready for RF-DETR's dynamic-shape requirements yet. The `onnx2tf` path in this PR is the pragmatically correct choice for now.

I'm tracking `litert-torch#870` and plan to revisit a migration once the dynamic-shape bug is resolved upstream. When that lands, this PR gives us a working baseline to migrate *from*, which is itself useful.

Thanks again for doing the research and flagging the better long-term path — that's exactly the kind of contribution that shapes where the project goes.

PR #964, opened 2026-04-13
fix: interpolate PE in `load_pretrain_weights` for custom resolution

### LINKED ISSUE #960: `RFDETRLarge(resolution=...)` silently falls back to deprecated config because `load_pretrain_weights` lacks positional embedding interpolation
### Search before asking

### Bug

This is a follow-up to #948 and PR #956. PR #956 correctly fixes the architecture-level part of the resolution problem — `positional_encoding_size` is now derived from the user-supplied `resolution` at construction time. 

However, while testing the merged fix on the `develop` branch, I found that the resolution path is still broken end-to-end, and the failure mode on `RFDETRLarge` is particularly subtle because a `try/except` in `RFDETRLarge.__init__` silently swallows the real error and falls back to the deprecated Large config. Users who pass a custom resolution to `RFDETRLarge` could end up training a completely different model.

Below are two connected problems with minimal reproductions.

### Environment

- RF-DETR: develop @ `c80bda9` (post-#956 merge)
- OS: Linux
- Python: 3.11
- PyTorch: 2.6.0+cu124
- GPU: NVIDIA GeForce RTX 3060

## Problem A — `load_pretrain_weights` lacks positional embedding interpolation

This is the cleanest reproduction. Default-construct the model (so the architecture matches the pretrained checkpoint), then pass `resolution=640` to `train()`:

```python
model = RFDETRLarge()
model.train(
    dataset_dir=str(DATASET_DIR),
    epochs=1,
    batch_size=8,
    grad_accum_steps=2,
    resolution=640,
)
```

```
RuntimeError: Error(s) in loading state_dict for LWDETR:
    size mismatch for backbone.0.encoder.encoder.embeddings.position_embeddings:
    copying a param with shape torch.Size([1, 1937, 384]) from checkpoint,
    the shape in current model is torch.Size([1, 1601, 384]).
```

What happens:

1. `RFDETRLarge()` constructs cleanly with the Large config (patch=16, PE=44, 704×704).
2. `train(resolution=640)` applies the new `resolution` to `model_config`, which (correctly, post-#956) rebuilds the model with PE=40 (1601 tokens).
3. `load_pretrain_weights` then tries to load the pretrained checkpoint (1937 tokens, 44×44) into the new model (1601 tokens, 40×40) and crashes on `load_state_dict`.

The same crash reproduces on `RFDETRSmall(resolution=640)` (1025 vs 1601) and `RFDETRMedium(resolution=640)` (1297 vs 1601), so it's not Large-specific — it's a general gap in `load_pretrain_weights`, which never interpolates positional embeddings between source and target grid sizes.

This is a standard ViT/DINOv2 operation: bicubic-interpolate the checkpoint's PE grid from the source shape (e.g. 44×44) to the target shape (e.g. 40×40) before calling `load_state_dict`. @isaacrob already confirmed in #956 that PE interpolation is acceptable as long as it's not baked into the export graph (interpolate, then trace).

Without this, the user-facing `resolution` parameter is unusable on any pretrained model whose checkpoint PE grid doesn't already match the target.

---

## Problem B — `RFDETRLarge` silently falls back to the deprecated config

This is a more subtle and arguably more dangerous failure. Construct `RFDETRLarge` with a custom resolution and then call `train()` with no resolution argument:

```python
model = RFDETRLarge(resolution=640)
model.train(
    dataset_dir=str(DATASET_DIR),
    epochs=1,
    batch_size=8,
    grad_accum_steps=2,
)
```

```
AssertionError: Backbone requires input shape to be divisible by 56,
but got torch.Size([8, 3, 640, 640])
```

The magic number `56` is the smoking gun. `56 = patch_size × num_windows`. For the **modern** Large config that's `16 × 2 = 32`. For the **deprecated** Large config it's `14 × 4 = 56`. So this error proves the model is being built with the deprecated Large architecture, even though the user wrote `RFDETRLarge(resolution=640)`.

The cause is in `src/rfdetr/variants.py`. `RFDETRLarge.__init__` wraps the modern-config construction in a try/except that falls back to the deprecated config if a specific set of error conditions is met:

```python
class RFDETRLarge(RFDETR):
    @staticmethod
    def _should_fallback_to_deprecated_config(exc: Exception) -> bool:
        message = str(exc).lower()
        if "out of memory" in message:
            return False
        if isinstance(exc, ValueError):
            return "patch_size" in message
        if isinstance(exc, RuntimeError):
            incompatible_state_dict_markers = (
                "error(s) in loading state_dict",
                "size mismatch",
                "missing key(s) in state_dict",
                "unexpected key(s) in state_dict",
            )
            return any(marker in message for marker in incompatible_state_dict_markers)
        return False

    def __init__(self, **kwargs):
        self.init_error = None
        self.is_deprecated = False
        try:
            super().__init__(**kwargs)
        except (ValueError, RuntimeError) as exc:
            if not self._should_fallback_to_deprecated_config(exc):
                raise
            self.init_error = exc
            self.is_deprecated = True
            try:
                super().__init__(**kwargs)
                logger.warning("... Automatically switched to deprecated model configuration ...")
            except Exception:
                raise self.init_error
```

The problem is that **the heuristic assumes the only cause of a state_dict size mismatch is deprecated weights**. That assumption held before #956. After #956, there is a second way to get a size mismatch: the user passes a custom `resolution`, the modern config correctly rebuilds the model with a new positional encoding grid (e.g. PE=40 for `resolution=640`), and then `load_pretrain_weights` crashes loading the default *modern* pretrained checkpoint (PE=44 for 704×704) because positional embeddings aren't interpolated (Problem A above).

Trace for `RFDETRLarge(resolution=640)` on current develop:

1. `super().__init__` tries the modern config (patch=16, num_windows=2).
2. #956 correctly sets `positional_encoding_size = 640 // 16 = 40`. Model is built with 1601 PE tokens.
3. `load_pretrain_weights` attempts to load the default `rf-detr-large.pth` (1937 tokens, 44×44) into the new 1601-token model.
4. PyTorch raises `RuntimeError: Error(s) in loading state_dict ... size mismatch for ... position_embeddings: [1, 1937, 384] vs [1, 1601, 384]`.
5. `_should_fallback_to_deprecated_config` inspects the error: it's a `RuntimeError` containing both `"error(s) in loading state_dict"` and `"size mismatch"` → returns `True`.
6. `is_deprecated = True`, and `super().__init__` is retried — this time `get_model_config` returns `RFDETRLargeDeprecatedConfig` (patch=14, num_windows=4, PE=37).
7. The deprecated Large path happens to succeed because its default pretrained weights *do* fit its own PE grid.
8. A warning is logged (*"Automatically switched to deprecated model configuration, due to using deprecated weights"*) — but the message is misleading in this scenario, because the user didn't pass any deprecated weights. They passed a custom resolution on a modern model, and the new size-mismatch path from #956 was misclassified as "deprecated weights".
9. Construction returns successfully. `self.is_deprecated = True` but the user has no reason to check that flag — they wrote `RFDETRLarge(resolution=640)` and expected modern Large.
10. Later, during training, the backbone's stride check fires: deprecated Large has `patch_size × num_windows = 56`, and `640 % 56 ≠ 0`, so the forward pass throws the AssertionError above.

The user sees a stride-related error referencing a number (`56`) that has no obvious connection to anything they did. The warning about deprecated configuration was logged earlier, but it's easy to miss and its phrasing doesn't match what the user actually did.

I confirmed the fallback path is the one being taken with a second experiment — forcing `patch_size=16` explicitly:

```python
model = RFDETRLarge(resolution=640, patch_size=16)
```

```
RuntimeError: size mismatch for ... position_embeddings: [1, 1937, 384] vs [1, 1601, 384]
```

With `patch_size=16` pinned, the deprecated fallback would produce a `ValueError` with `"patch_size"` in the message during retry (deprecated Large requires 14), which the inner `except Exception: raise self.init_error` then unwraps — so we end up seeing the real underlying error from Problem A. That confirms the PE mismatch *is* what trips the fallback in the unforced case.

### Why the heuristic is now wrong

`_should_fallback_to_deprecated_config` was written under an assumption that held before #956: any `state_dict` size mismatch means the user is trying to load deprecated-format weights into a modern model. Post-#956, that assumption no longer holds. There is now a legitimate reason for a size mismatch that has nothing to do with deprecated weights — namely, a custom `resolution` that changes the modern model's PE grid relative to the modern pretrained checkpoint.

The heuristic can't distinguish between these two cases from the error message alone, because both produce the exact same `"size mismatch for ... position_embeddings"` `RuntimeError`. The only reliable way to tell them apart is to inspect *why* the target model has a different PE grid: if it's because the user passed a custom `resolution`, the weights aren't deprecated — they're the correct modern weights, and they just need to be interpolated to the new grid size.

### Proposed fix

The cleanest fix is to solve Problem A — once `load_pretrain_weights` does PE interpolation, custom resolutions on modern Large no longer cause a size mismatch, the heuristic stops firing incorrectly, and the deprecated fallback only triggers in its intended scenario (actually-deprecated weights).

A more defensive fix would additionally tighten `_should_fallback_to_deprecated_config` so that a size mismatch triggers fallback only when the user did *not* pass a custom `resolution` (or, more robustly, only when the checkpoint itself is identified as deprecated-format via metadata). That way, even if PE interpolation is added later, the fallback can't silently hijack unrelated errors.

Either way, when the fallback *does* fire, it would be much easier to debug if the log message included the captured `init_error` so users can see *why* the fallback was chosen, not just *that* it happened.

### Additional

_No response_

- [ ] Yes, I'd like to help by submitting a PR!

Borda:
This pull request introduces robust support for custom input resolutions in the DINOv2 backbone by automatically resizing (interpolating) positional embeddings from pre-trained checkpoints to match the model's configured resolution. It also ensures that errors caused by explicit custom resolutions are not mistakenly attributed to deprecated weights, and adds comprehensive regression tests for these behaviors.

**Automatic positional embedding interpolation for custom resolutions:**

* Added `_interpolate_position_embeddings` to `weights.py`, which bicubic-interpolates positional embeddings from the checkpoint to match the model's configured grid size, preventing shape mismatch errors when loading weights with custom `resolution` settings.
* Integrated the new interpolation logic into `load_pretrain_weights`, ensuring all checkpoint loads correctly adapt positional embeddings before calling `load_state_dict`.

**Deprecation fallback improvements:**

* Updated the fallback logic in the model variant loader to avoid retrying with deprecated weights when a custom `resolution` is explicitly set, ensuring that genuine shape mismatches are surfaced as errors rather than being silently handled.

**Regression and unit tests:**

* Added targeted regression tests in `test_load_pretrain_weights.py` to verify that positional embeddings are interpolated correctly for various source and target resolutions, and that no modification occurs when the shapes already match.
* Added a regression test in `test_detr_shim.py` to ensure that specifying a custom resolution does not trigger the deprecated-config fallback on positional embedding size mismatches.

**Imports and dependencies:**

* Added necessary imports (`math`, `torch.nn.functional as F`) to support the new interpolation logic in `weights.py`.

---

Fixes #960

### DIFF
--- a/src/rfdetr/models/weights.py
+++ b/src/rfdetr/models/weights.py
@@ -42,9 +42,9 @@ _PE_KEY_SUFFIX = "embeddings.position_embeddings"
 
 def _interpolate_position_embeddings(
     checkpoint_state: dict,
-    model_config: ModelConfig,
+    pe_size: int,
 ) -> None:
-    """Interpolate DINOv2 positional embeddings in *checkpoint_state* to match *model_config*.
+    """Interpolate DINOv2 positional embeddings in *checkpoint_state* to match *pe_size*.
 
     When the model is configured with a custom ``resolution`` that differs from the
     checkpoint's training resolution, the DINOv2 backbone's ``position_embeddings``
@@ -52,16 +52,15 @@ def _interpolate_position_embeddings(
     skip shape mismatches on matching keys — it raises ``RuntimeError``.
 
     This function bicubic-interpolates every PE tensor in the checkpoint whose shape
-    differs from the target grid derived from ``model_config.positional_encoding_size``,
-    modifying *checkpoint_state* in-place before ``load_state_dict`` is called.
+    differs from the target grid, modifying *checkpoint_state* in-place before
+    ``load_state_dict`` is called.
 
     Args:
         checkpoint_state: The ``"model"`` sub-dict from a loaded checkpoint.
-        model_config: The model configuration used to derive the target PE size.
-            ``model_config.positional_encoding_size`` gives the target grid side length
-            (number of patches per spatial dimension, assuming a square grid).
+        pe_size: Target grid side length in patches (number of patches per spatial
+            dimension, assuming a square grid).  Typically
+            ``model_config.positional_encoding_size``.
     """
-    pe_size = model_config.positional_encoding_size
     n_target = pe_size * pe_size  # target number of patch tokens
 
     pe_keys = [k for k in checkpoint_state if k.endswith(_PE_KEY_SUFFIX)]
@@ -75,10 +74,8 @@ def _interpolate_position_embeddings(
         h_tgt = int(math.isqrt(n_target))
         if h_src * h_src != n_source or h_tgt * h_tgt != n_target:
             logger.warning(
-                "Skipping PE interpolation for %s: grid size is not a perfect square (source %d, target %d).",
-                key,
-                n_source,
-                n_target,
+                f"Skipping PE interpolation for {key}:"
+                f" grid size is not a perfect square (source {n_source}, target {n_target}).",
             )
             continue
 
@@ -262,7 +259,7 @@ def load_pretrain_weights(
         if any(name.endswith(x) for x in query_param_names):
             checkpoint["model"][name] = checkpoint["model"][name][:num_desired_queries]
 
-    _interpolate_position_embeddings(checkpoint["model"], mc)
+    _interpolate_position_embeddings(checkpoint["model"], mc.positional_encoding_size)
     nn_model.load_state_dict(checkpoint["model"], strict=False)
 
     # If the user explicitly set a class count larger than the checkpoint,
--- a/src/rfdetr/models/weights.py
+++ b/src/rfdetr/models/weights.py
@@ -19,11 +19,13 @@ extraction from ``detr.py:_load_pretrain_weights_into``.
 from __future__ import annotations
 
 import functools
+import math
 import os
 import warnings
 from typing import List
 
 import torch
+import torch.nn.functional as F  # noqa: N812
 
 from rfdetr.assets.model_weights import download_pretrain_weights, validate_pretrain_weights
 from rfdetr.config import ModelConfig, TrainConfig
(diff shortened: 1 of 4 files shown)


Alarmod:
Run RFDETRLarge with patch_size=16, num_windows=4, resolution=768 and pretrained weights - OK, train start without errors.

Another patch_size don't supported (patch_size=12).

Borda:
> Another patch_size don't supported (patch_size=12)

Let's create new issue for that, pls 🦝

PR #971, opened 2026-04-15
fix: infer `patch_size` from projection weight in `validate_checkpoint_compatibility`

### LINKED ISSUE #965: patch_size change for pretrained RF-DETR
### Search before asking

### Bug

I tested the code after the change: [https://github.com/roboflow/rf-detr/pull/964](https://github.com/roboflow/rf-detr/pull/964), but if I change the patch_size parameter in the RFDETRLarge constructor, explicitly specifying 12 instead of 16, the program breaks with the error
```
RuntimeError: Error(s) in loading state_dict for LWDETR:
size mismatch for backbone.0.encoder.encoder.embeddings.patch_embeddings.projection.weight: copying a param with shape torch.Size([384, 3, 16, 16]) from the checkpoint, the shape in the current model is torch.Size([384, 3, 12, 12])
```

### Environment

RF-DETR pre 1.6.5 (develop)
OS Windows 10 22H2
Python 3.13
PyTorch 2.11
GPU/CPU: Nvidia RTX 5070 Ti/Intel Core i5 12600k

```
model = RFDETRLarge(patch_size=12, num_windows=4, resolution=768)
model.train(...);
```

### Additional

_No response_

- [ ] Yes, I'd like to help by submitting a PR!

Borda:
This pull request improves the robustness and clarit#965  patch size compatibility checks between model checkpoints and model configurations, particularly addressing cases where pretrained weights lack explicit configuration metadata. It also adds comprehensive tests to ensure correct and backward-compatible behavior.

**Patch size inference and validation improvements:**

* Added logic in `validate_checkpoint_compatibility` to infer `patch_size` from the shape of the patch embedding projection weight when the checkpoint lacks an `"args"` key, ensuring early and clear errors for mismatched patch sizes and improving user experience when loading COCO-pretrained weights.

**Testing and backward compatibility:**

* Added parameterized tests to verify that patch size mismatches detected via projection weight inference raise clear errors, and that matching sizes do not raise.
* Added tests to confirm that the patch size inference logic is silently skipped when required keys or attributes are missing, preserving backward compatibility.
* Added a test to ensure that non-square projection kernels (where patch size cannot be reliably inferred) are skipped, preventing false errors for future or non-standard backbones.

---

resolves #965

### DIFF
--- a/src/rfdetr/utilities/state_dict.py
+++ b/src/rfdetr/utilities/state_dict.py
@@ -255,6 +255,26 @@ def validate_checkpoint_compatibility(checkpoint: dict[str, Any], model_args: An
                     ckpt_num_classes - 1,
                 )
 
+    # Infer patch_size from the patch-embedding projection weight when the checkpoint
+    # has no "args" key (e.g., COCO pretrained release weights that only store "model").
+    # Conv2d projection shape is [out_channels, in_channels, kernel_h, kernel_w];
+    # kernel_h == patch_size for square kernels.  Raises before load_state_dict fires,
+    # replacing the otherwise-cryptic "size mismatch" RuntimeError. Regression: #965.
+    # NOTE: key path is DINOv2-specific; non-DINOv2 backbones simply won't have this key
+    # and the check is silently skipped, preserving backward compatibility.
+    _patch_proj_key = "backbone.0.encoder.encoder.embeddings.patch_embeddings.projection.weight"
+    _ckpt_proj_w = checkpoint.get("model", {}).get(_patch_proj_key)
+    if _ckpt_proj_w is not None and _ckpt_proj_w.shape[2] == _ckpt_proj_w.shape[3]:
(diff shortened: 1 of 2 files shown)


Alarmod:
@Borda Now I get not just one warning, but two at once: 
 model = RFDETRLarge(patch_size=12, num_windows=2, resolution=704, num_queries=60, num_select=20, device=str(device))
 
```
ValueError: The checkpoint was trained with patch_size=14, but the current model uses patch_size=12. The checkpoint is incompatible with this model architecture. To resolve this, either instantiate/configure the model with the checkpoint's patch_size or use a checkpoint that was trained with the same patch_size as the current model.

During handling of the above exception, another exception occurred:

ValueError: The checkpoint was trained with patch_size=16, but the current model uses patch_size=12. The checkpoint is incompatible with this model architecture. To resolve this, either instantiate/configure the model with the checkpoint's patch_size or use a checkpoint that was trained with the same patch_size as the current model.
```

PR #1037, opened 2026-05-12
warn and disable training resize-and-crop branch when `aug_config={}`

### LINKED ISSUE #1034: Option to disable cropping and resizing during training
### Search before asking

- [x] I have searched the RF-DETR [issues](https://github.com/roboflow/rf-detr/issues) and found no similar feature requests.

### Description

Hi!

We are looking to implement RF-DETR as part of a model training pipeline for defect detection. When exploring [the code](https://github.com/roboflow/rf-detr/blob/develop/src/rfdetr/datasets/coco.py#L310) and documentation we have found that a standard resize and crop is applied on the input data. This cannot be disabled by setting `aug_config={}`.

The issue we are facing here is that we classify whole objects based on their defect. This cropping step risks removing the defect from the image, thus introducing noisy data samples. We already have a custom augmentation step that is equipped to handle this issue.

Would it be possible to provide a simple configuration parameter that allows us to disable this default behaviour?

Thanks in advance for considering this!

### Use case

_No response_

### Additional

_No response_

- [ ] Yes I'd like to help by submitting a PR!

### LINKED ISSUE #1175: Propose adding an Augmentation Cooldown Callback to Trainer
### Search before asking

- [x] I have searched the RF-DETR [issues](https://github.com/roboflow/rf-detr/issues) and found no similar feature requests.

### Description

Hey maintainers! 👋

I'd like to propose adding an **Augmentation Cooldown Callback** (`AugmentationCooldownCallback`) to the training pipeline. I have already coded this up, ran local tests (12 unit tests, 100% coverage), and formatted it to pass all pre-commit hooks. 

You can check out the implementation on my fork here:
👉 **Code Fork:** [venk-ee/rf-detr-ken](https://github.com/venk-ee/rf-detr-ken)

#### API Design
I designed this to align with the config patterns Borda mentioned in PR #1037—keeping it clean and avoiding adding unnecessary boolean flags:
* `aug_cooldown_config = None` -> Defaults to a mild baseline preset (`AUG_CONSERVATIVE` - mild flips/brightness).
* `aug_cooldown_config = {}` -> Completely disables augmentations during the cooldown.

#### Under the Hood Details (Why these are here)
* **Pydantic Validation**: Added a non-negative constraint (`ge=0`) to `aug_cooldown_epochs` in `TrainConfig` to prevent invalid negative values from breaking the trigger logic at startup.
* **Dynamic Rebuilds**: Dynamically rebuilds either the Kornia (GPU) or Albumentations (CPU) pipelines on the fly. This swaps the active augmentations midway through training without needing to restart the run.
* **Double Normalization Fix**: Explicitly forwards `gpu_postprocess = gpu_active` during CPU dataset rebuild. This ensures that when Kornia GPU training is active, the rebuilt CPU transforms skip normalization, preventing double-normalization (once on CPU, once on GPU) which would corrupt pixel values.
* **Keypoints Safety**: Preserves empty lists `[]` for `keypoint_flip_pairs` instead of letting it coerce to `None`. In Albumentations, `None` treats it as a detection pipeline, whereas `[]` keeps the wrapper in keypoints mode (preventing joint labels and coordinates from getting corrupted during horizontal flips).
* **Multiprocessing Warning**: Checks the fully resolved `_num_workers` and `_persistent_workers` from `datamodule` rather than raw config fields. This ensures we accurately warn users when the PyTorch worker cache prevents new transforms from propagating (which happens when `persistent_workers` defaults to `None` but resolves to `True`). The warning is suppressed on GPU runs (`not gpu_active`) since CPU worker caching has no impact on GPU Kornia pipelines.

**Would love to get your thoughts on this and open a PR!**

### Use case

**Why do this?**

During DETR training (especially with large backbones like DINOv2), heavy augmentations are crucial to prevent overfitting early on. However, keeping those extreme augmentations active until the very end can prevent the bounding box coordinates and loss from converging smoothly. 

Cooling down the augmentations in the final epochs gives the model a chance to stabilize and fine-tune on clean, realistic images. 

i used to use this in mm detection taught to add this into rf-detr

### Additional

* **100% Test Coverage:** I added a comprehensive unit test suite `tests/training/callbacks/test_augmentation_cooldown.py` (12 tests covering boundaries, CPU/GPU paths, keypoint structures, and warnings).
* **Pre-commit checks:** The code has been formatted, linted, and fully passes `pre-commit run --all-files` (ruff, mypy, docformatter, etc.).

- [x] Yes I'd like to help by submitting a PR!

omkar-334:
**Title:** warn and disable training resize-and-crop branch when `aug_config={}`

closes #1175

---

## What does this PR do?

Setting `aug_config={}` does not stop the default random resize-and-crop applied to training images. This is structurally correct but counter-intuitive: `aug_config` and the resize/crop logic live in two separate stages of the training pipeline, and `aug_config` only controls one of them.

In `make_coco_transforms` (`src/rfdetr/datasets/coco.py`), the training pipeline composes two stages:

1. **Resize stage.** Built from `_build_train_resize_config(...)`. It hard-codes a `OneOf` between:
    - Option A: direct resize to the target scale.
    - Option B: `SmallestMaxSize([400, 500, 600])`, then `RandomCrop`/`RandomSizedCrop`, then resize to the target scale.
2. **Augmentation stage.** Built from `aug_config` (color jitter, flips, rotations, etc.).

`aug_config={}` zeroes out stage 2 only. Stage 1 still runs, and the `RandomCrop`/`RandomSizedCrop` inside Option B is what clips whole objects out of frame in the reported defect-detection setup.

## How this PR fixes it

`aug_config` becomes the single knob that also governs the stage 1 crop branch. No new argument or config field is added.

| `aug_config` | Augmentations | Training resize-and-crop branch |
| --- | --- | --- |
| `None` (default) | default `AUG_CONFIG` preset | kept |
| `{}` | none | dropped, with a warning |
| non-empty dict | the given transforms | kept |

`None` is unchanged from current behavior. The only behavior change is `{}`: it already meant "no augmentations", and it now also drops the resize-and-crop branch so training uses direct resize only. A warning is logged on `{}` so the change is not silent.

Implementation:

- `_crop_branch_enabled(aug_config)` maps `aug_config` to a boolean and owns the `{}` warning.
- `_build_train_resize_config(..., include_crop_branch: bool)` stays a pure config builder. When `include_crop_branch=False` it returns Option A only, dropping the resize, crop, resize branch.
- `make_coco_transforms` and `make_coco_transforms_square_div_64` derive `include_crop_branch` from `aug_config` via `_crop_branch_enabled` and forward it.

Fixes #1034

## Usage

```python
from rfdetr import RFDETRMedium

model = RFDETRMedium()

# Default: augmentations on, resize-and-crop branch on
model.train(dataset_dir="path/to/dataset")

# Disable augmentations AND the resize-and-crop branch (logs a warning)
model.train(dataset_dir="path/to/dataset", aug_config={})
```

## Type of Change

## Testing

**Test details:**

- `tests/datasets/test_coco.py::TestAugConfigDisablesCrop`
    - `test_crop_branch_enabled` (parametrized over `None` / `{}` / non-empty) checks `_crop_branch_enabled` maps `aug_config` to the right decision.
    - `test_empty_aug_config_warns` checks `aug_config={}` logs a warning.
    - `test_make_coco_transforms_forwards_crop_decision` checks `make_coco_transforms` forwards the derived value to `_build_train_resize_config`.
- `tests/datasets/test_coco_resize_config.py::TestBuildTrainResizeConfigCropBranch::test_include_crop_branch_false_drops_crop_branch` (parametrized over square / non-square) checks `RandomCrop`/`RandomSizedCrop` are absent when `include_crop_branch=False`.

Existing characterization tests in `TestBuildTrainResizeConfigStructure` cover the default kept-branch case.
## Checklist

- [] I have updated the documentation accordingly (if applicable)


Borda (changes requested):
I see your point, but I'm not much in favor of adding a new argument. I would rather have `aug_config=None` to preserve actual structural behavior and None being the default, but enable forced reset with pasing `aug_config={}`.

omkar-334:
> I see your point, but I'm not much in favor of adding a new argument. I would rather have `aug_config=None` to preserve actual structural behavior and None being the default, but enable forced reset with pasing `aug_config={}`.

Do you mean if `aug_config` is `None`, then `_build_train_resize_config` uses that and disables the resize&crop?  basicaly reusing the same parameters?  
or do you not want this functionality at all

Borda:
> Do you mean if `aug_config` is `None`, then `_build_train_resize_config` uses that and disables the resize&crop? basicaly reusing the same parameters?

Yes 🦝, but if you have another suggestion, happy to hear it...

omkar-334:
I originally thought of that approach itself, but there are 2 things we should consider  

1. It merges two pipeline stages (resize-time geometry vs. augmentation-time pixel ops) into one knob. A user who wants `{}` augs but keeps the multi-scale resize+crop now has no way to express that.
2. Changing what `aug_config={}` can be a behavior break for anyone currently passing {} and relying on crop staying on.

what do you think?

Borda:
> * It merges two pipeline stages (resize-time geometry vs. augmentation-time pixel ops) into one knob. A user who wants `{}` augs but keeps the multi-scale resize+crop now has no way to express that.

We may move these multi-scale resize+crop to predefined augmentation, but need to check how to enforce/select which would use the receptive field size... so maybe really have these in predefined pipelines and in the model just add scaling to the model size as mandatory to not break the pre-processing

> * Changing what `aug_config={}` can be a behavior break for anyone currently passing {} and relying on crop staying on.

we can add some warning for people passing {}, which is use choce, not a default

cc: @isaacrob

omkar-334:
@Borda I removed the `do_random_crop` argument; `aug_config` now controls the resize-crop branch.

| `aug_config`     | Augmentations               | Train resize-crop branch |
| ---------------- | --------------------------- | ------------------------ |
| `None` (default) | default `AUG_CONFIG` preset | kept                     |
| `{}`             | none                        | dropped + warning        |
| non-empty dict   | the given transforms        | kept                     |

`None` is unchanged from current behavior. The only behavior change is `{}`: it already meant "no augmentations", and it now also drops the training resize-crop branch (direct resize only). A warning is logged on `{}` so that drop is not silent.

Borda:
> I removed the `do_random_crop` argument

I still see it...

omkar-334:
> > I removed the `do_random_crop` argument
> 
> I still see it...

I removed the public-facing parameter. We still need an internal flag for enabling this right... should i rename it?

omkar-334:
@Borda what do you think about the changes i made

Borda:
> Borda what do you think about the changes i made

a bit busy with next release, but will check when I have a moment... 🦝

Borda:
> > > I removed the `do_random_crop` argument
> > 
> > I still see it...
> 
> I removed the public-facing parameter. We still need an internal flag for enabling this right... should i rename it?

Is the internal param/arg needed?

docs/learn/train/advanced.md
```diff
@@ -285,6 +285,10 @@ To disable all augmentations, pass an empty dict:
 model.train(dataset_dir="path/to/dataset", aug_config={})
 ```
 
+Passing `{}` also drops the training resize-and-crop branch, so images are resized
```
Borda: lets format it as !!!Tip

Borda (approved):

PR #1109, opened 2026-06-15
fix: honor explicit `num_classes` even when it equals the default

anatoly-ryabchenko:
## What does this PR do?

Both train-time guards that decide whether the user overrode num_classes keyed off `("num_classes" in model_fields_set) and (num_classes != default)`. The `!= default` clause made the default value inconsistent: an explicit num_classes=90 (the class default) was silently realigned to the dataset's class count with no warning, while any non-default value was preserved and warned about.

Drop the clause at both guard sites (detr._align_num_classes_from_dataset and weights.load_pretrain_weights) so an explicit num_classes is always honored regardless of whether it equals the default; the unset case still auto-aligns to the dataset.

Follow-up to #1106 (approved as a separate PR in review). With #1106 merged, from_checkpoint no longer records checkpoint-derived num_classes in model_fields_set, so the `!= default` heuristic is no longer needed.

**Related Issue(s):** #1106

## Type of Change

- Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Testing

**Test details:**
- `tests/training/test_detr_shim.py::TestRFDETRTrainNumClassesAutoDetect::test_preserves_explicit_default_num_classes_when_dataset_differs`
  — *replaces* `test_auto_adjusts_when_default_explicitly_passed`. Covers the `train()`-time
  guard (`_align_num_classes_from_dataset`): passing `num_classes=90` (the class default)
  explicitly on a 4-class dataset now **preserves 90 and logs a warning** — identical to the
  non-default case, instead of silently realigning to 4.

- `tests/models/test_from_checkpoint.py::TestFromCheckpointNumClassesProvenance::test_explicit_default_num_classes_pins_head`
  — *replaces* `test_explicit_default_num_classes_does_not_block_alignment`. Covers a
  `from_checkpoint` model with an explicit default `num_classes`: the head is **pinned** (not
  silently realigned to the dataset) and a warning is logged. Explicitly guards against
  re-introducing the `!= default` clause.

- `tests/models/test_weights.py::TestLoadPretrainWeightsReinitScenarios::test_explicit_default_num_classes_treated_as_explicit`
  — *new*. Covers the second guard site (`load_pretrain_weights`): an explicit default
  `num_classes` + a smaller checkpoint reinitializes the head `[3, 91]` (load at the checkpoint
  size, then expand back to the configured size) rather than auto-aligning down to the
  checkpoint's class count.

## Checklist

## Additional Context

I flagged it as breaking change, since previously this edge case (`num_classes explicitly` set but equal to default) was handled differently - it was described both in docstrings and tests.

### DIFF
--- a/src/rfdetr/detr.py
+++ b/src/rfdetr/detr.py
@@ -1202,10 +1202,10 @@ class RFDETR:
         Must be called before ``RFDETRModelModule`` is constructed so that weight loading inside the module uses the
         correct (dataset-derived) class count.
 
-        When the user did **not** explicitly override ``num_classes`` (or passed the class-config default),
-        ``model_config.num_classes`` and ``self.model.args.num_classes`` are updated to match the dataset.  When the
-        user *did* set a non-default value that differs from the dataset, the configured value is preserved and a
-        warning is emitted.
+        When the user did **not** explicitly set ``num_classes`` (it is left unset, e.g. inferred from a
+        checkpoint), ``model_config.num_classes`` and ``self.model.args.num_classes`` are updated to match the dataset.
+        When the user *did* set ``num_classes`` explicitly — to any value, including the class default — and it differs
+        from the dataset, the configured value is preserved and a warning is emitted.
 
         Failures from ``_detect_num_classes_for_training`` are caught and logged at DEBUG level so that training is
         never blocked by detection errors.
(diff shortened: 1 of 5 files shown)


anatoly-ryabchenko:
@Borda will try to be more careful with docs and naming from now on

Borda (approved):

Borda:
> Borda will try to be more careful with docs and naming from now on 

Not sure what you mean if shall be more careful? 🦝

anatoly-ryabchenko:
> > Borda will try to be more careful with docs and naming from now on 
> 
> Not sure what you mean if shall be more careful? 🦝 

I mean I will try to be more careful so that you don't hVe to update them later 😅

PR #1121, opened 2026-06-16
fix(callbacks): always print validation summary tables under progress bar (#1119)

### LINKED ISSUE #1119: v1.8.0.rc2:Question: no longer see the validation result table
I just tried the v1.8.0.rc2. Nothing wrong but i dont see the validation table appear after completion of each epoch anymore. Is it intentionally removed? or do i miss any hyperparameter to unhide it?

Aaryan562:
## What does this PR do?
Fixes a regression where the per-epoch validation summary tables (Overall + Per-class metrics) stopped appearing in the terminal after v1.8.0 for anyone training with a progress bar.

COCOEvalCallback produces two distinct kinds of terminal output at epoch end:
- the raw pycocotools / faster-coco-eval stdout dump (noisy, multi-line), and
- the curated Rich summary tables (the clean Overall + Per-class metrics view).

## Root cause
The duplicate-output suppression added in ef14e363 gated both of these on the same _has_progress_bar(trainer) check in _compute_and_log.  The result: with progress_bar="tqdm" or "rich", no summary tables; with the config default progress_bar=None.

## Fix
Decouple the two concerns:

- Keep suppressing the raw COCO stdout dump under an active progress bar (unchanged — handled in _compute_map_metric).
- Always render the summary tables at epoch end. 

Related Issue(s): Closes [#1119](https://github.com/roboflow/rf-detr/issues/1119)

## Type of Change
☑ Bug fix (non-breaking change that fixes an issue)
Testing
☑ I have tested this change locally
☑ I have added/updated tests for this change

## Test details:
- Split the prior test_progress_bar_suppresses_terminal_metric_summaries into two focused tests in tests/training/test_coco_eval_callback.py:
  - test_progress_bar_suppresses_raw_coco_dump — verifies the raw stdout dump is still suppressed under a progress bar   (unchanged behavior).
  - test_summary_tables_printed_even_with_progress_bar — regression guard: summary tables render even when a progress bar is attached.
- Confirmed the regression test fails on pre-fix code and passes with the fix.
- Full tests/training/test_coco_eval_callback.py suite: 77 passed; ruff check / ruff format --check clean.

### DIFF
--- a/src/rfdetr/training/callbacks/coco_eval.py
+++ b/src/rfdetr/training/callbacks/coco_eval.py
@@ -595,8 +595,11 @@ class COCOEvalCallback(Callback):
             metrics=metrics, pfx=pfx, split=split, pl_module=pl_module, ar_by_cid=ar_by_cid, f1_by_cid=f1_by_cid
         )
 
-        if not self._has_progress_bar(trainer):
-            self._print_metrics_tables(trainer, split, overall, per_class)
+        # Always render the curated Rich summary tables. The noisy raw
+        # pycocotools/faster-coco-eval stdout dump is suppressed under an active
+        # progress bar in _compute_map_metric; these summary tables are not
+        # duplicate output and were previously hidden as a side effect (#1119).
+        self._print_metrics_tables(trainer, split, overall, per_class)
         self._compute_and_log_keypoint_map(split, pl_module, trainer)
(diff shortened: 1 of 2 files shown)


Borda:
As mentioned in the issue, this is not a bug; it was intentional. Let's first clarify what the desired behavior is...

PR #1137, opened 2026-06-19
fix: fixed python version mismatch dependency overrides for packages …

hinogi:
…depending on >=3.13 like numpy

## What does this PR do?

```
❯ uv sync
  × No solution found when resolving dependencies for split (markers: python_full_version >= '3.13'):
  ╰─▶ Because ml-dtypes==0.5.1 depends on numpy{python_full_version >= '3.13'}>=2.1.0 and onnx==1.20.1 depends on ml-dtypes{python_full_version == '3.12.*'}==0.5.1, we can conclude that onnx==1.20.1 depends on numpy>=2.1.0.
      And because onnx2tf>=2.4.0 depends on numpy==1.26.4, we can conclude that onnx==1.20.1 and onnx2tf>=2.4.0 are incompatible.
      And because onnx2tf>=2.4.0 depends on onnx==1.20.1 and only the following versions of onnx2tf are available:
          onnx2tf<=2.4.0
          onnx2tf==2.4.1
          onnx2tf==2.4.2
      we can conclude that onnx2tf>=2.4.0 cannot be used.
      And because rfdetr[tflite] depends on onnx2tf>=2.4.0 and your project requires rfdetr[tflite], we can conclude that your project's requirements are unsatisfiable.

      hint: While the active Python version is 3.12, the resolution failed for other Python versions supported by your project. Consider limiting your project's supported Python versions using `requires-python`.
```

fixing this conundrum

**Related Issue(s):** 

## Type of Change

## Testing

- [ ] I have tested this change locally

**Test details:**

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code where necessary, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings or errors
- [ ] I have updated the documentation accordingly (if applicable)

### DIFF
--- a/.github/workflows/ci-deps-resolution.yml
+++ b/.github/workflows/ci-deps-resolution.yml
@@ -1,0 +1,51 @@
+name: Dependency resolution
+
+on:
+  push:
+    branches: ["main", "release/*", "develop", "feat/*"]
+    paths: ["pyproject.toml"]
+  pull_request:
+    paths: ["pyproject.toml"]
+
+permissions:
+  contents: read
+
+defaults:
+  run:
+    shell: bash
+
+concurrency:
+  group: deps-resolution-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+jobs:
+  tflite-resolution:
+    name: Resolve [tflite] extra — Python ${{ matrix.python-version }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    strategy:
+      fail-fast: false
+      matrix:
+        # tflite deps carry python_version>='3.12' markers; tensorflow has no 3.13 wheels yet.
+        # Extend matrix when tensorflow adds 3.13 support.
+        python-version: ["3.12"]
+    steps:
+      - name: 📥 Checkout the repository
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+
+      - name: 🐍 Install uv and set Python version ${{ matrix.python-version }}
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -158,6 +158,10 @@ rfdetr = "rfdetr.cli:main"
 override-dependencies = [
     "ml-dtypes==0.5.1; python_version >= '3.12' and python_version < '3.13'",
 ]
+environments = [
+    "python_version >= '3.12'",
```
Borda: why >= 3.12 when package supports >= 3.10

hinogi: Because that was actually the only way that there where no dependency solution errors.

hinogi: I tried 3.10 and got the same error as without setting it to 3.12. Could it be a platform thing? Or am I doing something wrong?

hinogi:
```
tflite = [
    "onnx2tf>=2.4.0,<3.0.0",
...
]
```

So uv suggests pretty much the opposite lower version boundary like

```
tflite = [
    "onnx2tf<=2.4.0,<3.0.0",
...
]
```

if that is preferable

.github/workflows/ci-deps-resolution.yml
```diff
@@ -0,0 +1,51 @@
+name: Dependency resolution
+
+on:
+  push:
+    branches: ["main", "release/*", "develop", "feat/*"]
```
Borda:
```suggestion
    branches: ["main", "release/*", "develop"]
```

Borda (dismissed):

PR #1139, opened 2026-06-19
refactor: added typings for modules with missing source

hinogi:
## What does this PR do?

Generated typings for modules that miss sourcecode so there is a stub available

**Related Issue(s):** 

## Type of Change

- Refactoring (no functional changes)

## Testing

- [ ] I have tested this change locally

**Test details:**

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code where necessary, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings or errors
- [ ] I have updated the documentation accordingly (if applicable)

### DIFF
--- a/typings/faster_coco_eval/__init__.pyi
+++ b/typings/faster_coco_eval/__init__.pyi
@@ -1,16 +1,24 @@
 from __future__ import annotations
-from faster_coco_eval.core import coco
-from faster_coco_eval.core.coco import COCO
-from faster_coco_eval.core import faster_eval_api as cocoeval
-from faster_coco_eval.core.faster_eval_api import COCOeval_faster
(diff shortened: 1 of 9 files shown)


Borda:
I don't think we shall keep typing out other packages. I would suggest fixing the typing in the original dependencies

hinogi:
> I don't think we shall keep typing out other packages. I would suggest fixing the typing in the original dependencies

Sure, it was more of a developer convenience because otherwise it would be a deep rabbit hole of missing things in other packages. That's why I made this a separate PR so it can be dismissed if unwanted.

Borda:
For development let's rather use groups not extras 🦝

PR #1142, opened 2026-06-20
feat(export): add ExecuTorch (.pte) export

anatoly-ryabchenko:
## What does this PR do?

Add RFDETR.export(format="executorch"): torch.export -> Edge dialect -> backend delegation, writing a .pte (no ONNX step). ExecuTorch export required tracing spatial_shapes_hw through the deformable-attention decoder so the whole detection graph is torch.export-clean (numerically identical to the tensor path). Additionally, adds the optional [executorch] extra and tests/export/test_executorch_export.py (argument/dependency validation, format="executorch" wiring, and a gated end-to-end parity check). 

Dynamic batch is not supported for ExecuTorch — export(format="executorch", dynamic_batch=True) raises NotImplementedError. torch.export keeps the batch axis symbolic, but ExecuTorch 1.3.1 can't carry it through lowering for RF-DETR, for two independent reasons: (1) the Edge-dialect verifier specializes the batch dim to the traced value — it calls len() on the first tensor of a Tensor[]-returning op (the deformable-attention value.split(...)), which guards the batch to the example size; and (2) even past that, the windowed-attention reshapes (where batch is multiplied by the window count, B↔4·B) lower to view_copy/et_view ops the runtime can't resize for a variable batch. A dynamic .pte would therefore run only at the traced batch and silently mis-detect every other size, so it's refused. 

Supported backends:

- xnnpack (portable CPU, fp32) - default; validated end-to-end vs eager (~1e-5).
- coreml (Apple Neural Engine/GPU, fp16) - validated by detections (fp16 raw diffs are expected, as with other CoreML exports). Required limiting forward pass to rank-5 tensors, only in export branch.
- qnn (Qualcomm Snapdragon HTP, fp16) - requires an ExecuTorch source build against the Qualcomm AI Engine Direct (QNN) SDK, so it is not pip-installable. "aten.bitwise_not.default", "aten.lt.Scalar", "aten.topk.default", "aten.max.dim" run on CPU - these are the index-returning ops of the two-stage query selection (max over classes -> top-300 proposals). On the HTP fp16 path they compute WRONG indices, so the wrong 300 proposals are selected and every detection collapses (confidences ~0) even though the backbone + decoder match eager to rel_rmse 3e-4. Even on CPU, fp16 raw diffs, similar to CoreML.

Sample image results:
executorch_visual_4col
seg_4col
kp_4col

[...]

PR #1144, opened 2026-06-20
Fix: Slice Rank-4 reference-points before inserting broadcast axes.

tlandeen:
## What does this PR do?

Fixes the 4-coordinate reference-box path in `MSDeformAttn.forward` so it no longer produces an **over-ranked (rank-6) `StridedSlice`** during TFLite export.

Previously the two reference-box slices were taken on the *broadcast-expanded* tensor:

```python
reference_points[:, :, None, :, None, :2]        # rank-6 slice
+ ... reference_points[:, :, None, :, None, 2:]   # rank-6 slice
```

This change slices at **rank 4** (`xy` / `wh`) *before* inserting the broadcast (`None`) axes. The result is **numerically identical** but keeps every slice within TFLite's native **1D–5D** `StridedSlice` kernel limit.

**Why it matters:** the rank-6 slice makes the exported model fail to load on a strict/mobile LiteRT runtime, with the symptom depending on the onnx2tf version:

| onnx2tf | unfixed op | strict/mobile runtime result |
|---|---|---|
| 1.x | native rank-6 `STRIDED_SLICE` | rejected at interpreter creation: *"StridedSlice op only supports 1D-5D input arrays"* |
| 2.4.x | `FlexStridedSlice` (×2) | fails to compile: requires `tensorflow-lite-select-tf-ops` (Flex), not bundled in the standalone LiteRT runtime |

Same root cause (an over-ranked slice), so this single change resolves both. Desktop `tf.lite.Interpreter` (full TF) tolerates both forms, which is why the failure only appears on-device.

**Related Issue(s):** 
PR at #1142 may address the same issue

## Type of Change

## Testing

**Test details:**

Exported TFLite (fp16) from the same converter **with and without** this change and compared the op graphs. The only structural difference:

```
FlexStridedSlice   2 -> 0    (removed)
STRIDED_SLICE     64 -> 66   (+2, native)
RESHAPE          130 -> 132  (+2, native)
```

i.e. the 2 rank-6 `FlexStridedSlice` ops are replaced by native `STRIDED_SLICE` + `RESHAPE`; everything else (incl. the 305 fp16 weight tensors) is identical, so the model becomes Flex-free.

On-device verification (LiteRT standalone runtime, Google Pixel):
- **With this fix:** model loads/compiles and runs (CPU/XNNPACK fallback).
- **Without this fix:** fails to compile — *"Node 1490 (FlexStridedSlice) failed to prepare -> Failed to compile model."*

The change is a slice-reordering that is mathematically identical to the original (the broadcast axes are independent of the sliced last axis), so no numerical behavior changes.

## Checklist

## Additional Context

The change includes an inline comment explaining the rank-6 -> Flex/strict-kernel issue so the non-obvious slice ordering isn't "simplified" back later.

Broader context: this surfaced while evaluating RF-DETR for mobile GPU/NPU deployment. The fix is what makes the TFLite export **loadable on a stock mobile runtime at all**; it is independent of (and does not solve) the separate fact that the deformable decoder's `GATHER_ND`/`TOPK_V2`/`FLOOR_MOD` ops aren't supported by the LiteRT GPU delegate (the model runs CPU-only on-device).

### DIFF
--- a/src/rfdetr/models/ops/modules/ms_deform_attn.py
+++ b/src/rfdetr/models/ops/modules/ms_deform_attn.py
@@ -150,9 +150,11 @@ class MSDeformAttn(nn.Module):
                 + sampling_offsets / offset_normalizer[None, None, None, :, None, :]
             )
         elif reference_points.shape[-1] == 4:
+            reference_points_xy = reference_points[..., :2]
+            reference_points_wh = reference_points[..., 2:]
             sampling_locations = (
-                reference_points[:, :, None, :, None, :2]
-                + sampling_offsets / self.n_points * reference_points[:, :, None, :, None, 2:] * 0.5
+                reference_points_xy[:, :, None, :, None, :]
+                + sampling_offsets / self.n_points * reference_points_wh[:, :, None, :, None, :] * 0.5
             )
         else:
             raise ValueError(
--- a/src/rfdetr/models/ops/modules/ms_deform_attn.py
+++ b/src/rfdetr/models/ops/modules/ms_deform_attn.py
@@ -150,6 +150,13 @@ class MSDeformAttn(nn.Module):
                 + sampling_offsets / offset_normalizer[None, None, None, :, None, :]
             )
         elif reference_points.shape[-1] == 4:
+            # Slice the rank-4 reference_points BEFORE inserting the broadcast
+            # (None) axes.  Slicing the last axis of the rank-6 expanded tensor
+            # (reference_points[:, :, None, :, None, :2]) emits a rank-6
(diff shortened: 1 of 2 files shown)


Borda:
> Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/roboflow/rf-detr?pullRequest=1144) before we can accept your contribution.

@tlandeen pls ^^ otherwise we cannot land it 🦝

Borda:
Seems this was a fire and forgot PR... 
Since the SLA was not signed, unfortunately, there's nothing I can do/help to merge it... 🦝

Issue #1024, opened 2026-05-04
Expand model export format support


Borda:
# Expand model export format support

> **Updated 2026-07-30** — re-audited against `develop` @ `c3294a5` (version `1.9.0`). Since this issue was opened, the TFLite runtime failure has been fixed, TensorRT has been promoted to a first-class `format=`, and ExecuTorch + native CoreML have landed. Original text preserved where still accurate; per-format sections now cite the PR that added the capability and the release that shipped it.

## Context

RF-DETR now exposes five first-class export formats through `RFDETR.export(format=...)`. This issue audits the broader CV deployment ecosystem, records current coverage on `develop`, and tracks the remaining gaps.

Industry-share figures below are directional estimates, not precise telemetry — they aggregate signals from the Ultralytics export ecosystem, ONNX/TensorRT/CoreML deployment surveys, edge-AI framework reviews, and visible community usage.

## Current state in `develop`

Verified by inspection of `src/rfdetr/export/` and `RFDETR.export()` in `src/rfdetr/detr.py`:

- `RFDETR.export(format=...)` accepts `"onnx"` (default), `"tflite"`, `"tensorrt"` (alias `"trt"`), `"executorch"` (alias `"pte"`), and `"coreml"` (`detr.py:1431`).
- `src/rfdetr/export/` is now a package with per-backend subpackages: `_onnx/`, `_tflite/`, `_tensorrt.py`, `_executorch/`, `_coreml/`, plus a shared `ExporterProtocol` in `protocols.py` and dispatch in `_backend.py`.
- **The TFLite runtime failure flagged in the original issue is resolved.** Fixed by #1041 (onnx2tf 2.4) and #1054 (ONNX-level GridSample rewrite), both released in **1.7.0**; numerical parity with `predict()` closed by #1131 in **1.8.2**.
- TensorRT is no longer a `--tensorrt` CLI post-step shelling out to `trtexec` — it builds through the TensorRT Python API via `polygraphy` (#853, **1.9.0**).
- TorchScript remains in-process only via `optimize_for_inference(compile=True)` → `torch.jit.trace`. Not a saveable export format. Unchanged.
- Optional dependency groups in `pyproject.toml`: `[onnx]`, `[tensorrt]`, `[tensorrt-bench]`, `[tflite]`, `[executorch]`, `[coreml]`.
- Unreleased on `develop`: `output_name` + unified artifact naming across all backends (#1250).

## Overview

[...]

Issue #1233, opened 2026-07-24
ExecuTorch XNNPACK `.pte` returns 0 detections on real images


Borda:
**ExecuTorch XNNPACK `.pte` returns 0 detections on real images (backbone diverges from eager)**

## Summary

`RFDETRSmall().export(format="executorch", backend="xnnpack")` produces a `.pte` whose inference **diverges massively from eager PyTorch on real images**, collapsing every detection (all confidences < 0.25, all one class) → **0 detections**. It looks correct only because the existing parity test feeds **random** input, where the divergence does not appear.

- `predict()` on `dog.jpeg`: 4 detections `[0.95, 0.66, 0.58, 0.53]`
- eager **export-mode** forward, same image: same 4 detections `[0.96, 0.63, 0.58, 0.52]`
- XNNPACK `.pte`, same image: `[0.247, 0.241, 0.222, 0.208, 0.193]`, labels `[1,1,1,1,1]` → **0 above 0.5**

## Where the bug is (bisected)

`torch.export`'s ExportedProgram is **bit-exact** with eager; only the executorch-lowered `.pte` is wrong:

| Comparison | max abs delta logits |
| --- | --- |
| eager vs `torch.export` program (pre-lowering) | **0.0** |
| eager vs XNNPACK `.pte` | **10.47** |

Stage-by-stage inside the `.pte` (real image), divergence starts at the **DINOv2 windowed-attention backbone** and cascades:

| stage | shape | max abs delta (eager vs pte) |
| --- | --- | --- |
| backbone output | 1x256x32x32 | **3.09** (origin) |
| two-stage memory | 1x300x256 | 3.49 |
| decoder | 1x300x256 | 5.40 |
| pred_logits | 1x300x91 | 10.47 |

Ruled out:
- **`topk`/`max.dim`**: standalone through XNNPACK on the real scores gives 300/300 exact match (so the QNN-style index-divergence fix does not apply to XNNPACK).
- **`torch.export`**: perfect (0.0).
- **Regression**: reproduced at the #1142 merge commit too — present since the feature landed.

Input-dependence (the reason the current test misses it):

| input | max abs delta logits |
| --- | --- |
| `torch.randn` | 3e-4 |
| `torch.ones*0.5` | 1e-4 |
| real image (`dog.jpeg`, ImageNet-normalized) | **10.47** |

[...]

PR #1235, opened 2026-07-24
feat: Add CoreML export support and enhance related documentation

congbrian:
## What does this PR do?

Enables export to CoreML format on Apple Silicon hardware to close the 'largest deployable surface gap' identified in #1024. 

- New `format="coreml"` on `RFDETR.export()` which produces a native `.mlpackage` (mlprogram, iOS 16+) via `torch.export` + `coremltools`. No ONNX intermediary. Directly usable in CoreML/Xcode. Distinct from ExecuTorch's `format="executorch", backend="coreml" path`, which produces a .pte file.
- Builds on the `torch.export` groundwork from #1142 and #1229 
- New `rfdetr/export/_coreml/` package: converter, an op-coverage checklist (`unsupported_coreml_ops`) that flags registry gaps up front instead of failing during `ct.convert`, and patches to coremltools' Torch-op registry (`alias`, `__and__`, `bitwise_not`) that its registry currently lacks for RF-DETR's exported graph.
- Defaults to `compute_precision=FLOAT32` for tight CPU parity with eager PyTorch; `FLOAT16` is available on `export_coreml()` directly for smaller ANE-oriented bundles.
- New optional extra `rfdetr[coreml]` (`coremltools>=8.0,<10.0` — ExportedProgram support added in 8.0; validated on 9.0). Nothing CoreML-related is imported unless `format="coreml"` is requested.

**Related Issue(s):** #1024 

## Type of Change

## Testing

Test suite clears with all tests green. Added 11 tests for CoreML, in addition to tests on develop. Additionally ran smoke test with export + manual visualization of results.

**Test details:**
- `tests/export/test_coreml_export.py`: 
  - dependency/argument behavior with `coremltools` mocked (install-hint `ImportError`, `dynamic_batch` → `NotImplementedError`)
  - `RFDETR.export()` wiring (CoreML dispatch never touches the ONNX path and vice versa, notes warns, seg models dispatch, invalid format raises)
  - gated end-to-end tests (@pytest.mark.coreml, skipped when coremltools is absent) that run real exports of RFDETRNano and RFDETRSegNano and assert FLOAT32 CPU numerical parity against eager export-mode PyTorch (max abs diff < 1e-4 on boxes/logits/masks).
- tests/export/test_coreml_op_coverage.py
  - each registry patch is exercised and proven load-bearing (Nano's decomposed graph still emits `aten.alias`; the scanner flags `__and__` when unpatched)
  - Nano must be registry-clean after patches, enforced by an empty living allowlist so any new gap fails before merge. All registry mutations are snapshot/restored via an autouse fixture so a failing assert can't leak global state.
- New coreml pytest marker registered; existing export dispatch suites (test_export.py, test_executorch_export.py) pass unchanged.

## Checklist

## Additional Context
- Experimental, matching TFLite/ExecuTorch: a UserWarning on use, and dynamic_batch is rejected.
- Validated locally on macOS (Apple silicon) with coremltools 9.0 and torch 2.12. coremltools officially tests only through torch 2.7, but the parity suite passes; the version floor and the "validated on" comment follow the `[executorch]` pin convention.
- CI impact is minimal: the e2e/op-coverage tests skip wherever `coremltools` isn't installed, so existing CPU/GPU workflows are unaffected.
- Attached E2E visualization; left side Pytorch, right side CoreML.
image


congbrian:
Split tests into two regimes -- 1 set of lite tests covering operations/registry, one set of heavy tests covering full conversion + parity with real image test parity as per #1233 

Precommit failures are from upstream; mypy on `src/rfdetr/evaluation/coco_eval.py` and `src/rfdetr/training/trainer.py:86`

congbrian:
Validated and applied fixes from Copilot

congbrian:
Modified copilot solution as it was overly-general and could lead to cases where exceptions were nonspecific. Retained same behavior and guard.

tests/export/test_coreml_export.py
```diff
@@ -0,0 +1,461 @@
...
+    Returns:
+        Float tensor shaped ``(1, 3, resolution, resolution)``.
+
+    Raises:
+        FileNotFoundError: If ``path`` does not exist.
+    """
+    if not path.is_file():
+        raise FileNotFoundError(f"parity fixture image not found: {path}")
+    image = Image.open(path).convert("RGB")
+    tensor = TF.to_tensor(image)
```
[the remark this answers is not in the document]
congbrian: Fixed via 1cfc20a

tests/export/fixtures/images/dog.jpg
Borda, line 1: Let's do not add new images and use supervison assets

congbrian: Makes sense to me, handled in: 755646c284442f73a004f83507e099edace0cd26

Borda:
Since we have landed the ExecuTOrch, which also has a CoreML backend, what is the advantage of adding it as a separate export format? https://rfdetr.roboflow.com/develop/learn/export/#coreml-backend-apple-neural-engine-fp16

src/rfdetr/export/_coreml/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+# ------------------------------------------------------------------------
+"""CoreML export: PyTorch ``torch.export`` -> ``.mlpackage`` via coremltools."""
+
+from rfdetr.export._coreml.converter import _check_coremltools_available, export_coreml
+from rfdetr.export._coreml.op_coverage import unsupported_coreml_ops
+from rfdetr.export._coreml.torch_ops import ensure_coreml_torch_op_patches
+
+try:
+    _check_coremltools_available()
+    _IS_COREMLTOOLS_AVAILABLE: bool = True
```
Borda: This could be a cyclic import since it is expected to be used in the package
Also, I would expect that `_check_coremltools_available` returns a bool

src/rfdetr/export/_coreml/converter.py
```diff
@@ -0,0 +1,157 @@
...
+from rfdetr.export._coreml.op_coverage import unsupported_coreml_ops
+from rfdetr.export._coreml.torch_ops import ensure_coreml_torch_op_patches
+from rfdetr.utilities.logger import get_logger
+
+if TYPE_CHECKING:
+    import coremltools as ct
+
+logger = get_logger()
+
+_INSTALL_HINT = "CoreML export requires `coremltools`. Install it with: pip install rfdetr[coreml]"
```
Borda: Maybe rather wrap it as part of the func so the `_check_coremltools_available` would have arg if to raise hint

src/rfdetr/export/_coreml/converter.py
```diff
@@ -0,0 +1,157 @@
...
+        ImportError: If ``coremltools`` is not installed.
+        NotImplementedError: If ``dynamic_batch`` is requested, or if the exported graph
+            contains op kinds missing from coremltools' Torch registry (fast-fail checklist).
+        RuntimeError: If ``torch.export`` or ``coremltools.convert`` fails.
+    """
+    if dynamic_batch:
+        raise NotImplementedError(
+            "CoreML export does not support dynamic_batch (fixed shapes are required for reliable "
+            "ANE / GPU scheduling). Export one .mlpackage per batch size instead."
+        )
```
Borda: lets do not add it since it it not supported

congbrian:
> Since we have landed the ExecuTOrch, which also has a CoreML backend, what is the advantage of adding it as a separate export format? https://rfdetr.roboflow.com/develop/learn/export/#coreml-backend-apple-neural-engine-fp16

CoreML and ExecuTorch output different formats for different consumers. Based my own research, there's additional overhead associated with running ExecuTorch on Apple products, requiring developers to use the ExecuTorch platform, and introducing a C++ dependency.

CoreML is directly targeted towards Apple users, on the other hand, who can directly drag in the outputted .mlpackage which can be directly ingested by Apple's frameworks. MacOS/iOS developers can drag it into Xcode and load it with a few lines of Swift via the native API. This is the path that most Apple-native developers will prefer to use, as it produces the least friction -- they avoid adding a C++ dependency, a third party runtime, and changing their build system.

src/rfdetr/export/_coreml/__init__.py
congbrian: Makes sense. Fixed in 09ee66de70e62fedbfc2b2610e243a3553b2da34

src/rfdetr/export/_coreml/converter.py
congbrian: Wrapped: af063285e044ee18ced4efdfc0145358775fc89a

src/rfdetr/export/_coreml/converter.py
congbrian: Shifted tests to prevent usage of dynamic batching at the export function and removed it from the _coreml code. 7bad81580b7fefbf194a07ff921e50f6b6d4144f

src/rfdetr/export/_coreml/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+"""
+
+try:
+    import coremltools  # noqa: F401
+
+    _IS_COREMLTOOLS_AVAILABLE = True
+except ImportError:
+    _IS_COREMLTOOLS_AVAILABLE = False
+
+__all__ = ["_IS_COREMLTOOLS_AVAILABLE"]
```
Borda:
```suggestion
```
Stating a private Constant as a public API is not a good pattern

src/rfdetr/export/_coreml/converter.py
```diff
@@ -0,0 +1,142 @@
...
+    import coremltools as ct
+
+logger = get_logger()
+
+
+def _check_coremltools_available(*, raise_error: bool = True) -> bool:
+    """Return whether ``coremltools`` is importable."""
+    try:
+        import coremltools  # noqa: F401
+    except Exception as exc:
```
Borda:
```suggestion
    if not _IS_COREMLTOOLS_AVAILABLE:
```

src/rfdetr/export/_coreml/converter.py
```diff
@@ -0,0 +1,142 @@
...
+                exported_program,
+                convert_to="mlprogram",
+                minimum_deployment_target=ct.target.iOS16,
+                compute_precision=compute_precision,
+            )
+    except (ImportError, NotImplementedError, ValueError):
+        raise
+    except Exception as exc:
+        logger.exception("CoreML export failed")
+        raise RuntimeError(f"CoreML export failed: {exc}") from exc
```
Borda, line 173: What import error is left since you already have/know _IS_COREMLTOOLS_AVAILABLE?

docs/learn/export.md
```diff
@@ -121,6 +127,9 @@ model.export(backbone_only=True)
 After running the export, you will find the following files in your output directory:
 
 - `inference_model.onnx` - The exported ONNX model (or `backbone_model.onnx` if `backbone_only=True`)
```
Borda, line 129: Rather separate PR but also have this as `{variant}.onnx`

docs/learn/export.md
```diff
@@ -606,6 +622,99 @@ outputs = method.execute([input_tensor])
...
+ONNX intermediary and no ExecuTorch runtime dependency — the lowest-friction path for Apple-native
+(iOS / macOS) developers.
+
+### Prerequisites
+
+```bash
+pip install "rfdetr[coreml]"
+```
+
+Requires macOS and `coremltools>=8.0,<10.0`.
```
Borda: Let's not specify it in the docs, just in package extras

Borda (approved):

PR #1245, opened 2026-07-28
Optimize GPU memory usage for box iou during loss calculation

sergii-bond:
## What does this PR do?

There was unnecessary computation which created a huge matrix in memory when calculating box iou. It caused unpredictable training OOM, depending on torch memory cache allocator trying to find memory to store such matrix. And it had to do it for encoder and every decoder layer.

## Type of Change

- optimization

## Testing

**Test details:**

## Checklist

### DIFF
--- a/src/rfdetr/utilities/box_ops.py
+++ b/src/rfdetr/utilities/box_ops.py
@@ -69,6 +69,59 @@ def box_iou(boxes1: Tensor, boxes2: Tensor) -> tuple[Tensor, Tensor]:
     return iou, union
 
 
+def elementwise_box_iou(boxes1: Tensor, boxes2: Tensor) -> tuple[Tensor, Tensor]:
+    """Compute IoU and union for pre-matched box pairs.
+
+    Unlike ``box_iou``, this avoids materializing the full NxN pairwise matrix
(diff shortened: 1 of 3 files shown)


src/rfdetr/utilities/box_ops.py
```diff
@@ -69,6 +69,59 @@ def box_iou(boxes1: Tensor, boxes2: Tensor) -> tuple[Tensor, Tensor]:
...
+
+    lt = torch.max(boxes1[:, :2], boxes2[:, :2])  # [N,2]
+    rb = torch.min(boxes1[:, 2:], boxes2[:, 2:])  # [N,2]
+
+    wh = (rb - lt).clamp(min=0)  # [N,2]
+    inter = wh[:, 0] * wh[:, 1]  # [N]
+
+    union = area1 + area2 - inter
+
+    eps = 1e-7
```
Borda, line 110: Would this depend on the floating type, like FP32 or FP16?

sergii-bond: yes, the resulting eps after rounding will be slightly different in fp16, but here is mirrors what was in already present `box_iou` function here.

PR #1249, opened 2026-07-29
feat(augment): support ToGray on the Kornia backend

### LINKED ISSUE #1227: Augmentation built-in name: suggest adding "ToGray" method
### Search before asking

- [x] I have searched the RF-DETR [issues](https://github.com/roboflow/rf-detr/issues) and found no similar feature requests.

### Description

Hello there,

I find some dataset after apply `ColorJitter`, if we apply another effect "ToGray" that convert RBG to grayscale image and keep 3 channels will help improve the training. But I haven't found it in the built-in augmentation effect list (both Kornia and Albumentation).
Could you consider adding augmentation effect "ToGray" into the list?

### Use case

_No response_

### Additional

_No response_

- [ ] Yes I'd like to help by submitting a PR!

adhavan18:
## Description

Closes #1227.

`ToGray` was requested as a built-in augmentation. It turns out it already works on one backend and not the other, so this is really a parity gap rather than a new feature.

The Albumentations backend resolves names dynamically (`getattr(alb, name, None)` in [transforms.py](src/rfdetr/datasets/transforms.py)), so `ToGray` has always resolved there. The Kornia backend resolves from a fixed `_REGISTRY` in [kornia_transforms.py](src/rfdetr/datasets/kornia_transforms.py), which did not list it. The same `aug_config` therefore behaves differently depending on which backend `augmentation_backend` resolves to:

```text
kornia ToGray -> ValueError: Unknown augmentation key 'ToGray' for Kornia GPU backend
albu   ToGray -> OK
```

That is easy to hit without noticing, because `"auto"` picks Kornia when CUDA is present and a CPU-installed environment resolves elsewhere, so the same config can work locally and fail on a GPU box.

This maps `ToGray` onto `K.RandomGrayscale`, which matches Albumentations' semantics: the image is converted to greyscale and kept at three channels, so it stays a drop-in for an RGB pipeline. The `p` parameter passes through, defaulting to `0.5` like the other builders.

Also adds the row to the Kornia equivalence table and the colour list in `aug_configs.py`, so the documented built-ins stay in step with the registry.

## Testing

Three tests in `tests/datasets/test_kornia_transforms.py`, CPU-only like the rest of that module:

- `test_to_gray_builds_random_grayscale`: the key maps onto `RandomGrayscale`.
- `test_to_gray_keeps_three_channels_and_greys`: with `p=1.0`, the output keeps its three-channel shape and the channels are equal, which is the property that makes it a drop-in.
- `test_to_gray_accepted_by_both_backends`: the same config builds on Kornia and on Albumentations, so the parity gap stays closed.

All three fail on `a700efc` with `ValueError: Unknown augmentation key 'ToGray'` and pass here.

```text
$ pytest tests/datasets/test_kornia_transforms.py tests/datasets/test_augmentations.py
215 passed

$ ruff check src/rfdetr/datasets/kornia_transforms.py src/rfdetr/datasets/aug_configs.py tests/datasets/test_kornia_transforms.py
All checks passed!

$ ruff format --check ...
3 files already formatted
```

Ran on Windows, CPU only, `kornia 0.8.3` / `albumentations 2.0.8` / `torch 2.13.0+cpu`. No GPU-only paths are touched, so no `@pytest.mark.gpu` is needed.

### DIFF
--- a/src/rfdetr/datasets/aug_configs.py
+++ b/src/rfdetr/datasets/aug_configs.py
@@ -79,7 +79,7 @@ or torchvision defaults. Install it with ``pip install 'rfdetr[augment]'``.
 | ``Rotate`` | ``K.RandomRotation`` | ``limit`` may be scalar or tuple |
 | ``Affine`` | ``K.RandomAffine`` | ``translate_percent`` treated as fraction |
 | ``ColorJitter`` | ``K.ColorJiggle`` | Same multiplicative semantics |
-| ``ToGray`` | ``K.RandomGrayscale`` | Grayscale, kept at three channels |
+| ``ToGray`` | ``K.RandomGrayscale`` | Grayscale, 3 channels; only ``p`` honored, method/num_output_channels ignored |
 | ``RandomBrightnessContrast`` | ``K.ColorJiggle`` | ``brightness_limit`` / ``contrast_limit`` direct |
 | ``GaussianBlur`` | ``K.RandomGaussianBlur`` | ``blur_limit`` rounded up to odd; ``sigma=(0.1, 2.0)`` |
 | ``GaussNoise`` | ``K.RandomGaussianNoise`` | Upper bound of ``std_range`` used as fixed std |
--- a/src/rfdetr/datasets/aug_configs.py
+++ b/src/rfdetr/datasets/aug_configs.py
@@ -41,7 +41,7 @@ model.train(dataset_dir="...", aug_config={"HorizontalFlip": {"p": 0.5}})
 - Perspective: Perspective, ElasticTransform, GridDistortion
 
 **Pixel-level transforms** (preserve bounding boxes):
-- Color: ColorJitter, HueSaturationValue, RandomBrightnessContrast
+- Color: ColorJitter, HueSaturationValue, RandomBrightnessContrast, ToGray
 - Blur/Noise: GaussianBlur, GaussNoise, Blur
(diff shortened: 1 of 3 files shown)


adhavan18:
Note on the red `Testing` check: the failure is `tests/benchmarks/test_training_synthetic.py::test_train_convergence_segmentation`, a GPU training-convergence benchmark.

```text
E  AssertionError: val bbox mAP 0.062 should reach at least 0.15 after Trainer.fit.
E  assert 0.062261179089546204 >= 0.15
```

I do not think it is related to this PR:

- The test is already marked `@pytest.mark.flaky(reruns=1, only_rerun="AssertionError")`, so a stochastic convergence miss is a known mode for it.
- `develop` itself had a GPU tests Workflow failure at 2026-07-29T17:32 and a success at 17:33.
- #1250, which touches export naming and nothing in the augmentation path, is failing the same check right now.
- This PR only adds a new key to the Kornia `_REGISTRY` plus its builder. No existing code path changes, and no preset or benchmark config references `ToGray`, so the segmentation training run does not reach the new code at all.

The 38 other checks pass. Happy to rebase or push an empty commit if a re-run would help.

Borda:
@adhavan18, you are right, and it's annoying that it is not deterministic with some smaller variance. Would you be interested in working on stabilizing this test? 🦝

Borda (approved):

PR #1258, opened 2026-07-30
fix(docs): correct keypoint Params (126.4 → 40.7) and add Params to keypoint benchmarks

isaacrob:
## Summary

The keypoint model tables reported **126.4 M** parameters for `RFDETRKeypointPreview`. That is **RF-DETR-XL's** parameter count — apparently copied across because the keypoint checkpoint is named `rf-detr-keypoint-preview-xlarge.pth`. The correct value is **40.7 M**.

This PR corrects that number everywhere it appears, adds a `Params (M)` column to the keypoint benchmark comparison tables, and documents the accuracy methodology next to the existing latency note.

## Verifying the 40.7 M figure

Confirmed three independent ways:

| source | value |
| --- | --- |
| instantiated `nn.Module` | 40,724,248 (40.72 M) |
| released checkpoint tensors | 40,851,564 (40.85 M) |
| checkpoint file size (163.7 MB, all fp32) ÷ 4 | ~40.9 M |

The counting method (`sum(p.numel())`) was validated against the **already-published detection numbers** and reproduces them exactly, so the new figure is measured the same way as the rest of the table:

| model | measured | README |
| --- | --- | --- |
| `RFDETRNano` | 30.47 | 30.5 |
| `RFDETRSmall` | 32.11 | 32.1 |
| `RFDETRMedium` | 33.69 | 33.7 |
| `RFDETRLarge` | 33.93 | 33.9 |

## New Params column

Added to the keypoint comparison tables in `README.md` and `docs/learn/benchmarks.md`:

| | N | S | M | L | X |
| --- | --- | --- | --- | --- | --- |
| YOLO11-pose | 2.9 | 9.9 | 20.9 | 26.2 | 58.8 |
| YOLO26-pose | 2.9 | 10.4 | 21.5 | 25.9 | 57.6 |

RF-DETR Keypoint (Preview): **40.7**.

All YOLO values are **deployment (fused)** counts and match Ultralytics' published figures for all ten models (within ±0.06 M, i.e. rounding).

One footgun worth recording: raw *training-mode* counts for YOLO26-pose are much higher (3.68 M vs 2.9 M for nano; 62.73 M vs 57.6 M for X) because the YOLO26 head carries training-only structure — including a RealNVP `flow_model` — that folds away at deployment. `fuse()` is a silent no-op unless `eval()` is called first, which makes this easy to get wrong. YOLO11 is unaffected (raw ≈ fused), and RF-DETR contains **no BatchNorm at all**, so its fused and unfused counts are identical.

## Methodology note

The README previously documented only the *latency* methodology. Accuracy methodology was undocumented, which leaves readers with no explanation for why our third-party numbers differ from vendor-published ones. Added alongside the latency sentence:

> All COCO accuracy numbers are measured in-house for every model shown, computed with pycocotools in SAB over the full 5,000-image `val2017` split, so every row is directly comparable and may differ from vendor-reported figures.

`docs/learn/benchmarks.md` gets the equivalent in its existing **Methodology** section, plus a `Params (M)` definition in the metric-definitions block.

## Files changed

- `README.md` — keypoint model table (126.4 → 40.7), keypoint benchmark table (+Params), global methodology sentence
- `docs/index.md` — keypoint table (126.4 → 40.7)
- `docs/learn/run/keypoints.md` — keypoint table (126.4 → 40.7)
- `docs/learn/benchmarks.md` — keypoint benchmark table (+Params), Methodology paragraph, metric definitions

## Test plan

- [x] Param counts measured against latest `develop` (`c29743bb`)
- [x] Counting method validated against published detection numbers (exact match)
- [x] YOLO counts cross-checked against Ultralytics published values
- [x] `pre-commit` hooks pass on all changed files (`mdformat`, `codespell`, whitespace/EOF/line-ending)
- [x] All markdown tables verified for consistent column counts

Docs-only change; no code paths touched.


adhavan18 (commented):
Not a maintainer, but this is a numeric claim that can be checked independently, so I did.

Counted the parameters on a separate machine (Windows, CPU-only, `torch 2.13.0+cpu`), against the released checkpoint rather than your environment:

```python
from rfdetr import RFDETRKeypointPreview
m = RFDETRKeypointPreview()
net = m.model.model
sum(p.numel() for p in net.parameters())
```

```text
instantiated nn.Module params: 40,724,248  (40.72 M)
trainable                    : 40,724,248  (40.72 M)
```

That is your first row to the parameter, so 40.7 M reproduces cleanly and the 126.4 M figure is definitely wrong.

Two things I did not check, so treat them as unverified rather than confirmed:

- The attribution of 126.4 M to RF-DETR-XL. Your explanation is plausible given the `-xlarge.pth` filename, but I did not instantiate XL to confirm the number actually matches it.
- The AP and latency numbers in the tables, which need SAB and a T4.

One observation on the new `Params (M)` column: your note says parameter counts are deployment (fused) values, and the two counts in your table differ (40.72 M instantiated vs 40.85 M from checkpoint tensors, roughly 127 k apart, about what fused BN buffers would account for). Since the column is a single number, it might be worth stating in the note which of the two the column reports, so a reader reproducing it lands on the same figure rather than the other one.

Borda (approved):

Issue #1266, opened 2026-08-01
[onnx] fails to install on Python 3.10 and [executorch] on 3.14


Vedanshu7:
### Search before asking

- [x] I have searched the RF-DETR issues and found no similar bug report.

### Bug

Two optional extras cannot be installed on Python versions the project claims to support. `requires-python` is `>=3.10`, but `[onnx]` fails on 3.10 and `[executorch]` fails on 3.14.

Both fail the same way. The resolver picks a version, and then the install step finds there is no wheel for that interpreter. Neither package ships a source build, so there is no fallback.

```
$ uv sync --extra onnx --python 3.10
Resolved 373 packages in 249ms
error: Distribution `onnxruntime==1.24.3` can't be installed because it doesn't
have a source distribution or wheel for the current platform
hint: You're using CPython 3.10 (`cp310`), but `onnxruntime` (v1.24.3) only has
wheels with the following Python implementation tags: `cp311`, `cp312`, `cp313`, `cp314`
```

```
$ uv sync --extra executorch --python 3.14
error: Distribution `executorch==1.3.1` can't be installed because it doesn't
have a source distribution or wheel for the current platform
hint: You're using CPython 3.14 (`cp314`), but `executorch` (v1.3.1) only has
wheels with ABI tags: `cp310`, `cp311`, `cp312`, `cp313`
```

I checked both against the PyPI JSON API, so this is not specific to my machine:

* `onnxruntime` stopped shipping cp310 wheels after 1.23.2. Every release from 1.24.0 on covers cp311 to cp314 only. The entry in `[onnx]` is unpinned, so a 3.10 resolve picks the newest release and then cannot install it.
* `executorch` has never shipped a cp314 wheel. 1.3.1 is the current release and it stops at cp313.

The other eight extras (`tflite`, `coreml`, `augment`, `train`, `cli`, `loggers`, `visual`, `lora`) install fine on 3.10 through 3.14.

### Environment

```
RF-DETR:   1.9.0 (develop, f50258b0)
OS:        macOS 26.5.2, arm64
Python:    3.10.20 and 3.14.6 (the failing versions), both from uv
PyTorch:   2.9.1
CUDA:      none, CPU-only build
Hardware:  Apple M4 Pro
uv:        0.11.6
```

PyTorch, CUDA, and hardware are not involved here. The failure happens during dependency install, before any model code runs, and it is about wheel tags rather than the platform, so it should reproduce on Linux and Windows too.

[...]

PR #1267, opened 2026-08-01
fix(deps): gate onnx and executorch extras to interpreters with wheels

Vedanshu7:
## What does this PR do?

Issue: https://github.com/roboflow/rf-detr/issues/1266

Two optional extras cannot be installed on Python versions the project claims to support. `requires-python` is `>=3.10`, but `[onnx]` fails on 3.10 and `[executorch]` fails on 3.14.

Both fail the same way. The resolver picks a version, and then the install step finds that there is no wheel for that interpreter. Neither package ships a source build, so there is no fallback.

```
$ uv sync --extra onnx --python 3.10
Resolved 373 packages in 249ms
error: Distribution `onnxruntime==1.24.3` can't be installed because it doesn't
have a source distribution or wheel for the current platform
hint: You're using CPython 3.10 (`cp310`), but `onnxruntime` (v1.24.3) only has
wheels with the following Python implementation tags: `cp311`, `cp312`, `cp313`, `cp314`

$ uv sync --extra executorch --python 3.14
error: Distribution `executorch==1.3.1` can't be installed because it doesn't
have a source distribution or wheel for the current platform
hint: You're using CPython 3.14 (`cp314`), but `executorch` (v1.3.1) only has
wheels with ABI tags: `cp310`, `cp311`, `cp312`, `cp313`
```

Checked against PyPI, so this is not a local quirk:

* `onnxruntime` stopped shipping cp310 wheels after 1.23.2. Every release from 1.24.0 on covers cp311 to cp314 only. The entry was unpinned, so a 3.10 resolve picks the newest one.
* `executorch` has never shipped a cp314 wheel. 1.3.1 is the current release, and it stops at cp313.

### The fix

Add environment markers so each extra picks a version that has a wheel for the interpreter in use.

```toml
- "onnxruntime"
+ "onnxruntime<1.24; python_version < '3.11'"
+ "onnxruntime; python_version >= '3.11'"

- "executorch>=1.3,<2.0"
+ "executorch>=1.3,<2.0; python_version < '3.14'"
```

On 3.14, the `[executorch]` extra becomes installable but empty rather than raising an error. That is the same approach already used for `[tflite]`, which is gated to 3.12. Both entries carry a comment saying when the gate can be removed.

### Why CI did not catch this

`uv lock` only proves a version solution exists. It does not prove that the chosen version has a wheel for a given interpreter. I confirmed `uv lock --python 3.11` passes on the broken code, so CI stayed green while `uv sync` failed for real users.

The workflow file already noted the gap: "No CI job runs `uv sync`, so nothing else covers this path."

So this PR also adds an `install-plan` job to `ci-deps-resolution.yml`. It runs `uv sync --dry-run` for every extra on 3.10 through 3.14. Dry run does resolution plus install planning, which is where the wheel check happens, and it downloads nothing, so the job stays cheap.

One detail worth flagging for future changes: `uv pip install --dry-run` is not a substitute. It re-resolves per interpreter and quietly falls back to an older version that does have a wheel, which hides this exact bug. I tried that first, and it passed on the broken code. The job comment records this.

`tensorrt`, `tensorrt-bench`, `xla`, and `plus` are left out of the loop because they need NVIDIA or TPU indexes, or the separately licensed `rfdetr_plus` package, none of which resolve on a plain runner.

**Related Issue(s):** None. Found while installing the extras locally.

## Type of Change

## Testing

**Test details:**

No unit tests were added. This is a packaging bug, and a unit test cannot tell whether PyPI has a wheel for a given interpreter without doing the network work that the resolver already does. The regression check is the new CI job, which fails on the current code and passes with the fix.

What I ran:

| Check | Result |
| --- | --- |
| Reproduce on clean `develop` | `[onnx]` fails on 3.10, `[executorch]` fails on 3.14 |
| New CI job against current `develop` | Fails, and flags exactly those two, nothing else |
| New CI job with this fix | 10 extras across 5 Python versions, all pass |
| Real install (not dry run) on 3.10 | Resolves `onnxruntime==1.23.2` and installs |
| `pre-commit run --all-files` | All hooks pass except mypy, see below |
| CPU test suite, CI marker set | 3284 passed, 57 skipped, 0 failed |

Test command used:

```
uv run --no-sync pytest src/ tests/ -n 4 \
  -m "not gpu and not coco17 and not e2e_coreml and not e2e_executorch and not xla and not tpu" \
  --ignore=tests/run_smoke_all_models.py \
  --ignore=tests/legacy/test_checkpoint_compat.py --timeout=240
```

On mypy, it reports 45 errors on an unmodified `develop` checkout and the same 45 on this branch. They are all `import-not-found` for packages in the `[train]` extra, which `uv sync --all-groups` does not install. With those extras installed, the count drops to 2, again identical on both. This PR touches no Python files, so none of it comes from these changes.

## Checklist

On the last two: the new markers and the CI job both carry comments explaining the reason and the condition for removing them, matching the style already used for the `[tflite]` gate and the `[tool.uv]` conflicts. `AGENTS.md` listed the CI workflows but was missing `ci-deps-resolution.yml`, so I added an entry for it that also tells future contributors to add new extras to the job loop.

## Additional Context

`ci-tests-cpu.yml` does install packages on 3.10 through 3.13, but it would not have caught this for two reasons. It installs only `train,augment,cli,visual` plus `coreml` on macOS, so `onnx`, `executorch`, `tflite`, `loggers`, and `lora` have no install coverage at any version. It also installs with `uv pip install`, which backtracks past a missing wheel instead of failing. That is why the new job covers the full 3.10 to 3.14 range rather than just the two ends.

### DIFF
--- a/.github/workflows/ci-deps-resolution.yml
+++ b/.github/workflows/ci-deps-resolution.yml
@@ -115,3 +115,27 @@ jobs:
       - name: Minimize uv cache
         continue-on-error: true
         run: uv cache prune --ci
+
+  deps-resolution-complete:
+    name: Dependency resolution complete
+    # Single aggregate gate for branch protection: require THIS one check instead of the ~70
+    # install-plan matrix contexts (5 interpreters × every extra) plus the two resolution jobs.
+    # `needs.<job>.result` collapses a whole matrix to one value — `success` only if every cell
+    # passed. `if: always()` so the gate still runs (and fails) when an upstream job fails;
+    # without it a failed dependency would skip this job and branch protection would see no signal.
+    if: always()
+    needs: [universal-resolution, list-extras, install-plan]
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: ✅ Assert every resolution job succeeded
+        run: |
+          results="universal-resolution=${{ needs.universal-resolution.result }} list-extras=${{ needs.list-extras.result }} install-plan=${{ needs.install-plan.result }}"
+          echo "$results"
+          for pair in $results; do
+            if [ "${pair#*=}" != "success" ]; then
+              echo "::error::Dependency resolution incomplete — $results"
+              exit 1
+            fi
+          done
+          echo "All dependency-resolution jobs passed."
--- a/.github/workflows/ci-deps-resolution.yml
+++ b/.github/workflows/ci-deps-resolution.yml
@@ -81,18 +81,38 @@ jobs:
         continue-on-error: true
         run: uv cache prune --ci
 
-  install-plan:
-    name: Install plan for every extra — Python ${{ matrix.python-version }}
+  list-extras:
+    name: List extras for install-plan
     runs-on: ubuntu-latest
-    timeout-minutes: 15
+    timeout-minutes: 5
+    outputs:
+      extras: ${{ steps.collect.outputs.extras }}
+    steps:
+      - name: 📋 Collect the extras that get install-plan coverage
+        id: collect
+        # Single source of truth for the install-plan matrix below — `fromJSON` fans this list out
+        # into one job per extra, so there is no in-step loop to keep in sync. tensorrt /
+        # tensorrt-bench / xla / plus are excluded: they need NVIDIA/TPU indexes or the separately
+        # licensed rfdetr_plus package, none of which resolve on a plain runner. Add a new extra
+        # to this list to give it install-plan coverage.
+        run: |
+          echo 'extras=["onnx","executorch","tflite","coreml","augment","train","cli","loggers","visual","lora"]' >> "$GITHUB_OUTPUT"
+
+  install-plan:
+    name: Install plan [${{ matrix.extra }}] — Python ${{ matrix.python-version }}
+    needs: list-extras
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
     strategy:
       fail-fast: false
       matrix:
-        # Every interpreter allowed by requires-python (>=3.10). ci-tests-cpu.yml installs only
-        # train/augment/cli/visual (+coreml on macOS), so onnx, executorch, tflite, loggers and
-        # lora have no install coverage there at any version — and it installs via `uv pip
+        # Every interpreter allowed by requires-python (>=3.10) crossed with every extra from the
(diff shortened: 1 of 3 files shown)


Vedanshu7:
@Borda, the Socket alert here is from my change, not a false positive.

onnxruntime 1.23.2 needs coloredlogs, which pulls in pyreadline3 on Windows. 1.24.0 dropped coloredlogs, but 1.24 has no Python 3.10 wheel. So any version that installs on 3.10 brings the extra packages with it. No version avoids both.

It only affects Python 3.10, and pyreadline3 only lands on Windows. Socket gives 1.23.2 a clean score for known vulnerabilities.

I would keep it as is. 3.10 is tested on Linux, Windows and macOS in ci-tests-cpu.yml, and onnxruntime is the only blocker. onnx, onnxsim, onnx-graphsurgeon and polygraphy all still ship for 3.10. The pin also removes itself once requires-python moves past 3.10.

The counter-argument is that 3.10 goes end of life in October 2026. If you would rather not carry the pin, bumping requires-python to >=3.11 is cleaner than gating just this extra. That drops the pin, this alert and the 3.10 CI leg in one go.

So: keep as is, or bump the floor to 3.11? Happy to push either.

Also, CI has not run on this PR yet since it is my first contribution from a fork. Could you approve the workflow run when you get a chance?

.github/workflows/ci-deps-resolution.yml
```diff
@@ -72,10 +72,68 @@ jobs:
...
+        # failure mode that broke [onnx] on 3.10 and [executorch] on 3.14.
+        #
+        # `uv sync --dry-run` runs resolution AND install planning (where the wheel check
+        # happens) without downloading anything, so this stays cheap. `uv pip install --dry-run`
+        # is NOT equivalent: it re-resolves per interpreter and silently backtracks to an older
+        # version with a compatible wheel, hiding exactly this bug.
+        #
+        # tensorrt/tensorrt-bench/xla/plus are excluded: they need NVIDIA/TPU indexes or the
+        # separately licensed rfdetr_plus package, none of which resolve on a plain runner.
+        run: |
```
Borda: Let's rather refactor to have one simple job loading all these extras in a list, convert from JSON, and use them as parameters in startegy in the testing job so not have a loop inside the step but have each extra as a job

Borda (dismissed):

Borda (approved):

PR #1206, opened 2026-07-13
fix(inference): disable antialias in `predict()` resize to match training

Borda:
This pull request ensures that the image resizing behavior during inference matches the interpolation used during training, addressing a regression issue that caused discrepancies in bounding box and confidence values. The update disables antialiasing in the resize operation and adds a regression test to verify this behavior.

**Inference preprocessing consistency:**

* Updated `src/rfdetr/detr.py` so that `F.resize` in the `predict()` method is called with `antialias=False`, matching the antialias-free bilinear resize (cv2.INTER_LINEAR) used by Albumentations during training. This prevents drift in bounding box and confidence values relative to pretrained checkpoints.

**Testing:**

* Added a new test class `TestPredictResizeMatchesTrainingInterpolation` in `tests/inference/test_predict.py` to verify that `predict()` disables antialiasing during resize, ensuring inference preprocessing matches training-time interpolation.

---

resolves #1203

Issue #1203, opened 2026-07-13
Resize preprocessing inconsistencies across training/predict/exported models


ankandrew:
RF-DETR seems to use different resize paths:

- [Current CPU training/validation](https://github.com/roboflow/rf-detr/blob/e877b758c309ca749c739c477cb96f1668af0e9f/src/rfdetr/datasets/coco.py#L685-L705) uses Albumentations `Resize` w/o specifying interpolation (`INTER_LINEAR` by default).
- [`predict()`](https://github.com/roboflow/rf-detr/blob/e877b758c309ca749c739c477cb96f1668af0e9f/src/rfdetr/detr.py#L1925-L1953) uses torchvision `F.resize()` with its antialiased bilinear defaults.
- The [ONNX example](https://github.com/roboflow/rf-detr/blob/e877b758c309ca749c739c477cb96f1668af0e9f/docs/learn/export.md#L360-L373) uses Pillow `resize()` without specifying `resample`, which defaults to **bicubic**.

Also, [earlier training](https://github.com/roboflow/rf-detr/blob/bfe92571fbfb3f0000410b369c71db7e1ec6dd6e/src/rfdetr/datasets/coco.py#L194-L238) used [`SquareResize`](https://github.com/roboflow/rf-detr/blob/bfe92571fbfb3f0000410b369c71db7e1ec6dd6e/src/rfdetr/datasets/transforms.py#L229-L237), which called torchvision `F.resize()` on PIL images.

I find this confusing and inconsistent, and saw some bbox/conf values drift when porting to ONNX. Btw, which resize preprocessing was used for published pretrained ckpts?

JESUSROYETH:
Hi @ankandrew 👋

You are correct, the three paths do not match. I checked each one on current `develop`:

1. Training / val (what the pretrained checkpoints saw): Albumentations `Resize` / `SmallestMaxSize`, which default to `cv2.INTER_LINEAR` — bilinear, no antialias.
2. `predict()`: torchvision `F.resize(t, resize_to)` at `detr.py` L1952. In torchvision 0.27 this defaults to `BILINEAR` **with `antialias=True`**.
3. The ONNX docs example: `PIL Image.resize((w, h))` with no `resample`, and PIL defaults to **bicubic**.

So, the answer to your question: the published checkpoints were trained with Albumentations bilinear (`cv2.INTER_LINEAR`), antialias off. That is the reference to match at inference.

[...]

PR #1269, opened 2026-08-02
fix(export): match predict()'s antialias=False resize in ONNX/TFLite inference, benchmark and INT8 calibration

JESUSROYETH:
Follow-up to #1203 and #1206. #1206 switched `predict()` to `antialias=False` so inference matches the resize the released checkpoints were trained with, but the export-side preprocessing paths kept their old conventions, so exported models no longer see the same tensors as `predict()`:

1. TFLite inference (`_preprocess_image`) calls `F.resize` with no flag, and torchvision defaults to `antialias=True` for float tensors. This path was made bit-exact against the old `predict()` in #1131 and nobody updated it after #1206.
2. ONNX inference (`_preprocess_pil_to_nchw`) resizes with PIL BILINEAR, which applies an adaptive antialias filter when downscaling.
3. The ONNX/TensorRT benchmark and the export example-input pipeline resize the PIL image before tensorising, so they also go through PIL's antialiased filter.
4. INT8 calibration resizes with PIL's default resample, which is BICUBIC since Pillow 9.1.

I measured the divergence on 25 COCO images, comparing float32 bits through a `uint32` view (baseline first checked against itself: 0 bits differ):

| resolution | path | max abs diff (normalised space) | bits differing |
|---|---|---|---|
| 384 | TFLite current | 1.603 | 96.41% |
| 512 | TFLite current | 0.695 | 86.76% |
| 576 | TFLite current | 0.423 | 83.35% |
| 384 | ONNX current | 1.602 | 91.64% |

End to end it moves detections, not just pixels. Released RFDETRNano at 384, threshold 0.3, 10 COCO images: the canonical preprocessing gives 50 detections, the current TFLite preprocessing gives 47. Hungarian matching restricted to the same class leaves 47 pairs, mean IoU 0.951, and a max score difference of 0.153 — detections appear and disappear around the threshold just from the resize flag.

The parity test did not catch it because its baseline omits `antialias=False` too, and its "bit exact" assertion was a 1e-5 tolerance.

**Changes**

- Every export path now resizes with `predict()`'s exact convention: bilinear, half-pixel centres, `antialias=False`. With torchvision importable, TFLite and ONNX preprocessing are bit-identical to `predict()` — 0 differing bits on the same 25 images at the three resolutions.
- `_bilinear_resize_half_pixel` moved from `_tflite/inference.py` to a shared `rfdetr/export/_resize.py`, and it now also backs the torch-free fallback of both inference helpers. It implements the same antialias-free half-pixel convention, so the fallback gap drops from ~1.6 to ~5e-5 in normalised space. The old PIL BILINEAR fallback cannot get closer, its filter is antialiased by design.
- The benchmark and export example-input pipelines tensorise before resizing (`ToImage → ToDtype → Resize(antialias=False)`). The only remaining difference with `predict()` there is 1 ULP from v2's uint8→float scaling.
- INT8 calibration resizes with the same convention, so the calibration ranges come from the pixel distribution the model actually sees at inference. Existing INT8 exports keep working, and re-exporting picks up the aligned calibration.
- The TFLite parity test now asserts bit-identity (uint32 view, no tolerance) on the torchvision path, and a new test checks the antialiased variant stays far away, so the assertion really discriminates on the flag. The ONNX helper gets its own parity file (`test_onnx_preprocess_parity.py`), it had no preprocessing test before. Same for the two remaining consumers: `test_infer_preprocess_parity.py` locks `infer_transforms` and `make_infer_image` to `predict()`'s convention, and `test_tflite_export.py` adds a calibration-resize test. Each of these embeds a discriminator assertion (the old antialiased behaviour must stay far from the reference), so none of them can become vacuous like the previous parity test did.
- One sentence in the migration guide updated to reflect that `predict()` and the export inference paths are antialias-free since #1206.

**Validation**

- Every new or hardened test fails on current `develop` and passes with this change: the TFLite parity cases (389600/442368 bits differ on the 1280x720 → 384 case), the ONNX bitwise and fallback cases, both benchmark/example-input cases, and the calibration-resize case. Verified by reverting each touched source file individually with the tests in place.
- `pytest tests/export/` — 303 passed, 63 skipped, 0 failed.
- Full CPU suite — same result as `develop`, the only failures are pre-existing environment ones in `tests/cli` (missing `jsonargparse` locally), unrelated to this change.
- `pre-commit run` clean on the touched files; mypy on `src/rfdetr/export/` clean.

### DIFF
--- a/src/rfdetr/export/_onnx/inference.py
+++ b/src/rfdetr/export/_onnx/inference.py
@@ -11,6 +11,7 @@ RF-DETR training stack — only ``onnxruntime``, ``numpy``, ``supervision``, and
 
 from __future__ import annotations
 
+import contextlib
 from pathlib import Path
 from typing import Any
 
@@ -114,7 +115,7 @@ def _preprocess_pil_to_nchw(
     mean_list = [_imagenet_mean[i % 3] for i in range(channels)]
     std_list = [_imagenet_std[i % 3] for i in range(channels)]
 
-    try:
+    with contextlib.suppress(ImportError):
         # Match predict() exactly: torchvision to_tensor -> resize(antialias=False) -> normalize.
         # antialias=False mirrors detr.py's predict(); torchvision's float-tensor default is True.
         import torch
@@ -125,8 +126,6 @@ def _preprocess_pil_to_nchw(
             t = _F.resize(t, [height, width], antialias=False)
             t = _F.normalize(t, mean_list, std_list)
         return np.asarray(t.unsqueeze(0).cpu().numpy(), dtype=np.float32)
-    except ImportError:
-        pass
 
     # Torch-free fallback: same antialias-free half-pixel bilinear as predict(), in NumPy.
     arr = np.asarray(pil_img, dtype=np.float32) / 255.0
--- a/src/rfdetr/export/_onnx/inference.py
+++ b/src/rfdetr/export/_onnx/inference.py
@@ -152,11 +152,14 @@ def _run_inference(
 
     **Input contract** (must match ``RFDETR.predict()`` preprocessing exactly):
 
-    - Image is opened as-is and converted to ``"RGB"`` (3-channel) or ``"L"``
-      (1-channel greyscale) depending on the model's channel count.
-    - Resize uses ``PIL.Image.Resampling.BILINEAR`` — matching
-      ``torchvision.transforms.functional.resize()`` which defaults to ``InterpolationMode.BILINEAR``.  Using PIL's
-      default (``BICUBIC``) would produce slightly different pixel values and can degrade confidence.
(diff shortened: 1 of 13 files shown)


src/rfdetr/export/_onnx/inference.py
```diff
@@ -105,22 +109,34 @@ def _preprocess_pil_to_nchw(
...
-            image.convert(pil_mode).resize((width, height), PILImage.Resampling.BILINEAR),
-            dtype=np.float32,
-        )
-        / 255.0
-    )
+    pil_img = image.convert(pil_mode)
+    mean_list = [_imagenet_mean[i % 3] for i in range(channels)]
+    std_list = [_imagenet_std[i % 3] for i in range(channels)]
+
+    try:
```
Borda: Rather use context supress

Borda (approved):

Issue #1252, opened 2026-07-30
Kornia backend rejects 12 transform documented in `aug_configs`


adhavan18:
### Search before asking

- [x] I have searched the RF-DETR issues and found no similar bug report.

### Bug

Twelve of the transform names listed under **Transform Categories** in [`aug_configs.py`](https://github.com/roboflow/rf-detr/blob/develop/src/rfdetr/datasets/aug_configs.py) raise `ValueError` on the Kornia backend while working on Albumentations. Because `"cpu"`/`"auto"` resolves to Kornia when it is installed and CUDA is available, the same `aug_config` can train fine on a CPU box and hard-fail on a GPU box.

I built every documented name on both backends on `a700efc`:

| documented name | kornia | albumentations |
|---|---|---|
| `HorizontalFlip` | OK | OK |
| `VerticalFlip` | OK | OK |
| `Rotate` | OK | OK |
| `Affine` | OK | OK |
| `ShiftScaleRotate` | **ValueError** | OK |
| `RandomCrop` | **ValueError** | OK |
| `CenterCrop` | **ValueError** | OK |
| `RandomResizedCrop` | **ValueError** | OK |
| `Perspective` | **ValueError** | OK |
| `ElasticTransform` | **ValueError** | OK |
| `GridDistortion` | **ValueError** | OK |
| `ColorJitter` | OK | OK |
| `HueSaturationValue` | **ValueError** | OK |
| `RandomBrightnessContrast` | OK | OK |
| `GaussianBlur` | OK | OK |
| `GaussNoise` | OK | OK |
| `Blur` | **ValueError** | OK |
| `CLAHE` | **ValueError** | OK |
| `Sharpen` | **ValueError** | OK |
| `Equalize` | **ValueError** | OK |

To be fair to the docstring: the **Kornia GPU Backend** section right below lists exactly the eight that do work, and it is accurate. The mismatch is that the categories list above it reads as the menu of available transforms with no backend caveat, and the two lists disagree. All four shipped presets stay inside the supported eight, so nothing is broken out of the box.

The reason it bites is the failure mode. `AlbumentationsWrapper` resolves names dynamically (`getattr(alb, name, None)`), so it picks up anything Albumentations ships. `kornia_transforms._REGISTRY` is a fixed eight-entry dict, and an unlisted key is a hard `ValueError` at pipeline-build time, not a warning or a fallback.

#1227 is one instance of this (`ToGray`, PR #1249).

### Environment

[...]

PR #1277, opened 2026-08-04
feat(augment): support Blur, Sharpen, Equalize and CLAHE on the Kornia backend

adhavan18:
## Description

Makes progress on #1252.

That issue lists twelve documented transform names that raise `ValueError` on the Kornia backend while working on Albumentations. This PR closes four of them, the ones with faithful Kornia equivalents and no open design question:

| preset key | Kornia | note |
|---|---|---|
| `Blur` | `K.RandomBoxBlur` | Albumentations' `Blur` is a box blur, so this is a direct mapping |
| `Sharpen` | `K.RandomSharpness` | `alpha` is the blend weight, which is what Kornia's `sharpness` factor controls |
| `Equalize` | `K.RandomEqualize` | |
| `CLAHE` | `K.RandomClahe` | `clip_limit` and `tile_grid_size` map straight onto `clip_limit` and `grid_size` |

None of the four can move a box, mask or keypoint, so they sidestep the auxiliary-target question raised on the issue entirely. Before:

```text
Blur      kornia=ValueError  albumentations=OK
Sharpen   kornia=ValueError  albumentations=OK
Equalize  kornia=ValueError  albumentations=OK
CLAHE     kornia=ValueError  albumentations=OK
```

After, all four build on both backends.

Parameters Kornia cannot express are logged rather than dropped silently, matching how `ToGray` already handles `method` and `num_output_channels`. `Sharpen` warns about `lightness` and `method`; `Equalize` warns about `mode`, `by_channels` and `mask`.

## One change of scope from what I offered on the issue

On #1252 I said I would send five, including `HueSaturationValue` mapped to `K.ColorJiggle`. Looking at it properly, that mapping is not faithful and I have left it out.

Albumentations shifts hue, saturation and value **additively** (`hue_shift_limit=(-20, 20)` in OpenCV hue units, `sat_shift_limit=(-30, 30)` on a 0-255 scale). Kornia's `ColorJiggle` scales them **multiplicatively** by a factor. Mapping one onto the other would silently change what an existing config means rather than port it, so I would rather leave `HueSaturationValue` raising a clear `ValueError` than quietly do something different on the GPU path. There is a test pinning that it still raises, and the docs now say why.

The geometric group is untouched for the same reason as before: it needs the target-handling decision first.

## Refactor

`Blur` needs the same kernel resolution `GaussianBlur` already had (a `(min, max)` pair collapses to its upper bound, forced odd and at least 3), so that logic moved into a shared `_as_odd_kernel()` helper and `GaussianBlur` now calls it. Behaviour is unchanged and the existing `GaussianBlur` tests cover it, including the `blur_limit` pair cases.

## Testing

Thirteen cases added to `tests/datasets/test_kornia_transforms.py`:

- each of the four builds the expected Kornia class
- each is accepted by **both** backends, which is the parity gap the issue is about
- `Blur` kernel resolution is odd and at least 3 across scalars and pairs, warns on a lossy pair, stays quiet on a degenerate one
- `Sharpen` passes `alpha` through as a range and warns about `lightness`
- `Equalize` warns about ignored options
- `CLAHE` maps both parameters
- `HueSaturationValue` still raises, so the exclusion is deliberate rather than accidental

All thirteen fail on the parent commit with `ValueError: Unknown augmentation key`.

```
$ pytest tests/datasets/test_kornia_transforms.py tests/datasets/test_augmentations.py
248 passed
```

`ruff check` and `ruff format` are clean on the changed files. Tests are CPU-only, like the rest of that module.

### DIFF
--- a/src/rfdetr/datasets/aug_configs.py
+++ b/src/rfdetr/datasets/aug_configs.py
@@ -83,6 +83,16 @@ or torchvision defaults. Install it with ``pip install 'rfdetr[augment]'``.
 | ``RandomBrightnessContrast`` | ``K.ColorJiggle`` | ``brightness_limit`` / ``contrast_limit`` direct |
 | ``GaussianBlur`` | ``K.RandomGaussianBlur`` | ``blur_limit`` rounded up to odd; ``sigma=(0.1, 2.0)`` |
 | ``GaussNoise`` | ``K.RandomGaussianNoise`` | Upper bound of ``std_range`` used as fixed std |
+| ``Blur`` | ``K.RandomBoxBlur`` | Box blur; ``blur_limit`` rounded up to odd, pair collapses to its upper bound |
+| ``Sharpen`` | ``K.RandomSharpness`` | ``alpha`` maps to ``sharpness``; ``lightness``/``method`` ignored |
+| ``Equalize`` | ``K.RandomEqualize`` | Only ``p`` honored; ``mode``/``by_channels``/``mask`` ignored |
+| ``CLAHE`` | ``K.RandomClahe`` | ``clip_limit`` and ``tile_grid_size`` map directly |
+
+Not yet supported on Kornia: ``HueSaturationValue`` (Albumentations shifts hue/saturation/value additively,
+Kornia's ``ColorJiggle`` scales them multiplicatively, so there is no faithful mapping), and the geometric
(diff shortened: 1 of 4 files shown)


adhavan18:
Closing this for now. The branch stays on my fork, so I can reopen it if it is still wanted.

Borda:
> Closing this for now. The branch stays on my fork, so I can reopen it if it is still wanted.

What was your judgment to close it?

Borda (dismissed):

PR #1281, opened 2026-08-04
perf(matcher): copy only the diagonal cost blocks to CPU before assignment

JESUSROYETH:
## What does this PR do?

`HungarianMatcher.forward` copies the full `[batch, num_queries, total_targets]` cost matrix to CPU before calling `linear_sum_assignment`, then splits it per image and keeps only `c[i]`, the diagonal block for that image's own targets. Every column that belongs to another image in the batch gets copied to host and thrown away right after.

This PR extracts the diagonal blocks on-device first (using the cumulative `target_offsets` of each image's target count), and copies only the compact `[num_queries, total_targets]` matrix to CPU instead of the full `[batch, num_queries, total_targets]` one. So the host transfer drops by exactly the batch size on the normal (finite) path. The non-finite path is a bit different: `isfinite` gets checked on-device before any transfer, and if any cost is NaN/Inf the code still falls back to copying the *full* matrix to CPU and running the existing `_sanitize_cost_matrix`, same as before, that part I didn't touch. The diagonal-block extraction that runs afterwards (`target_offsets` + `torch.cat`, replacing the old `cost_matrix.split(sizes, -1)` + `c[i]`) is new code, but it's shared by both branches now instead of duplicated on each side of the `if`. So it doesn't add a new asymmetry between the finite and non-finite paths, it's just relocated.

Measured on real COCO batches, A100-SXM4-40GB, PyTorch 2.9.1+cu128, `RFDETRSmall` at 512 / `RFDETRSegSmall` at 384, batch 16, median of 31 interleaved repeats with warmup (ranges are min–max):

| flow, batch 16 | targets | before | after | change |
| --- | ---: | ---: | ---: | ---: |
| detection, COCO batch 0 | 111 | 45.321 ms (44.386–47.676) | 24.508 ms (24.199–27.420) | **45.9%** |
| detection, COCO batch 2 | 129 | 51.785 ms (50.610–53.274) | 27.750 ms (27.400–28.729) | **46.4%** |
| detection, COCO batch 233 | 205 | 74.014 ms (73.769–74.410) | 40.714 ms (40.528–41.259) | **45.0%** |
| segmentation, COCO batch 0 | 111 | 16.802 ms (16.120–18.512) | 13.853 ms (13.536–16.067) | 17.5% |

The criterion calls the matcher four times per step, so the effect adds up: full criterion on COCO batch 2, 31 reps, goes from 227.711 ms (225.719–234.298) to 129.176 ms (127.482–139.981), 43.3%. `forward + criterion + backward` (no `optimizer.step`), same batch, 9 reps, goes from 391.457 ms (386.566–399.230) to 289.659 ms (288.120–294.566), 26.0%.

CPU-only control (single thread, 11 reps) shows no clear regression, medians moved -0.7%, -9.5% and -3.2% for batches 1/4/16, but the ranges overlap so I'm not calling that an improvement either, just no regression.

**Related Issue(s):** none, this is a candidate I found while profiling the matcher, not a reported bug.

## Type of Change

- Performance improvement (no behaviour change)

## Testing

- During development I ran an ad hoc script (not part of this PR) comparing assignment indices before/after across 10 seeds for both detection and segmentation batches (1/4/16): bit-for-bit identical. That script isn't committed, so it's not independently re-runnable from this diff alone, see the two bullets below for what *is* committed and re-runnable.
- Added `TestDiagonalBlockExtraction` in `tests/models/test_matcher.py`, comparing the new `target_offsets` extraction against a reference that reproduces the previous `cost_matrix.split(sizes, -1)` + `c[i]` extraction, through the real `matcher.forward` on detection-shaped costs. Covers heterogeneous target counts with a zero-target image in the middle of the batch, `group_detr > 1`, and the degenerate case where every image in the batch has zero targets (all three cases assert equality against the reference, not just that the output is empty/well-shaped).
- Added `TestDiagonalExtractionContentAgnostic`, which isolates the extraction step from cost computation and property-tests the old and new extraction algorithms directly against 15 arbitrary cost tensors (5 seeds × `group_detr` in {1, 2, 3}). The mask (`cost_mask_ce`/`cost_mask_dice`) and keypoint (`cost_l1`/`cost_findable`/`cost_visible`/`cost_nll`) cost terms get summed into the same `cost_matrix` before this extraction runs (`matcher.py:266-277`, untouched by this PR), and the extraction has no notion of where its input came from. So this ends up being the committed regression test for segmentation and keypoint batches too, without re-deriving their cost formulas here.
- Exceptional paths (an injected NaN, and the whole cost matrix turned non-finite) still produce identical indices, these are the pre-existing `TestHungarianMatcherNonFiniteCosts`/`TestHungarianMatcherSanitization` tests, which exercise the shared extraction step end-to-end and still pass unmodified.
- The 17 criterion losses have identical bytes before/after, and gradients with respect to every captured model output have the same SHA-256 in both versions, checked with an ad hoc script during development (not committed, same caveat as above; both claims follow from the assignment-index equivalence above plus the fact that `criterion.py` is untouched). I couldn't verify bit-identity for every parameter gradient of the full model though, the CUDA baseline is not deterministic against itself (`grid_sampler_2d_backward_cuda` has no deterministic implementation in PyTorch), so that comparison stops at the boundary that matters here (criterion gradients w.r.t. outputs), not beyond it.
- CUDA peak memory is unchanged: the temporary cost matrices dominate the peak either way. The win is transfer time and volume, not VRAM.
- `tests/models/test_matcher.py` passes in full: 35 passed (17 existing + 3 new in `TestDiagonalBlockExtraction` + 15 new in `TestDiagonalExtractionContentAgnostic`).
- `ruff check` and `ruff format` are clean on the touched files, `mypy` finds no issues in `matcher.py`.

I didn't touch anything else, `HungarianMatcher` is only used from `criterion.py`, and there's no second copy of this splitting logic elsewhere in the codebase (unlike the `postprocess.py` case in #1265/#1268, which had a twin in `export/benchmark.py`).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -312,13 +312,12 @@ addopts = [
     "--color=yes",          # colored output
     "--doctest-plus",       # run doctests from all modules (pytest-doctestplus)
 ]
-# Skip doctest COLLECTION for test modules (they are not doc sources; src/ doctests unaffected).
-# Without this, doctestplus walks every test module's attributes; when the TF/ai_edge_litert
-# stack has been imported by the tflite export tests, scanning a module that exposes
-# `unittest.mock.call` (a `_Call` that auto-creates and caches child mocks on attribute access)
-# grows the object graph unboundedly — observed 38 GB RSS and a silent SIGKILL during
-# collection. The value must be a glob ("tests/*"): matching is per-file via fnmatch.
-doctest_norecursedirs = ["tests/*"]
+# Doctest collection is intentionally enabled across tests/ (no doctest_norecursedirs exclusion) so
+# helper-function doctests in test modules run. This previously blanket-excluded tests/* because
+# doctestplus walking every test module's attributes after the TF/ai_edge_litert stack is imported by
+# the tflite export tests grows unittest.mock.call's cached child-mock graph unboundedly — observed
+# 38 GB RSS and a silent SIGKILL during collection. If that recurs, reintroduce
+# `doctest_norecursedirs = ["tests/*"]` or scope it to `tests/export/test_tflite_inference.py`.
 pythonpath = ["src", "."]
 markers = [
     "gpu: tests that require GPU or are slow on CPU",
--- a/src/rfdetr/models/matcher.py
+++ b/src/rfdetr/models/matcher.py
@@ -264,7 +264,7 @@ class HungarianMatcher(nn.Module):
             cost_nll = cost_nll.flatten(0, 1)
 
         # Final cost matrix
-        cost_matrix = self.cost_bbox * cost_bbox + self.cost_class * cost_class + self.cost_giou * cost_giou
+        cost_matrix: Tensor = self.cost_bbox * cost_bbox + self.cost_class * cost_class + self.cost_giou * cost_giou
         if masks_present:
             cost_matrix = cost_matrix + self.cost_mask_ce * cost_mask_ce + self.cost_mask_dice * cost_mask_dice
         if keypoints_present:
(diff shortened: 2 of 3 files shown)


JESUSROYETH:
Hi @Borda 👋

Quick heads up on this one: I applied one of Copilot's suggestions here, moving the finite check to a scalar so the full-size mask doesn't stay alive on GPU, a pretty small change. When I applied it I dropped the indentation on the two new lines, so it broke every CI job with an IndentationError right at import. Already fixed with a follow-up commit, tests pass again.

No rush, but I have a follow-up PR in draft that depends on this one landing, just wanted it on your radar for whenever you get a chance to look ...

Borda (approved):

Issue #564, opened 2026-01-21
[Refactor]: Improve typing and sanitize docs


Borda:
### Search before asking

- [x] I have searched the RF-DETR [issues](https://github.com/roboflow/rf-detr/issues) and found no similar feature requests.

### Description

I suggest we remove them from the docstrings and rely solely on the signature. Here is the breakdown of why the Python ecosystem has shifted this way:

### 1. The Modern Approach (Signature as Truth)
Since PEP 484, the standard is to treat the function signature as the **Single Source of Truth**.
Tools like **mypy**, **Pyright**, and our IDEs rely entirely on the signature to validate our code. By keeping the focus there, we ensure that the types we document are the actual types being enforced by the interpreter.

### 2. The Legacy Approach (Why to avoid duplication)
Writing types in the docstring was necessary before Python 3.5, but today it is considered a violation of **DRY (Don't Repeat Yourself)**.
* **Maintenance Burden:** It creates a "split truth." If we change a variable from `int` to `float` in the code but forget to update the English text in the docstring, our documentation becomes factually incorrect.
* **Style Standards:** The **Google Python Style Guide** now explicitly states: *"The description should include the type if and only if the function does not have a type annotation."*

### 3. Do we need both for the docs?
No. You might be worried that removing them will break our generated documentation, but tools like **Sphinx** (via `sphinx-autodoc-typehints`) or **MkDocs** now automatically pull the types from the signature and insert them into the generated HTML. We get the same result without the manual effort.

### Are you willing to submit a PR?

- [x] Yes I'd like to help by submitting a PR!

Abdul-Mukit:
Are there any more scripts that we can take on?
I saw the migration plan to PyTorch Lightning. I was wondering if the MRs related to improving typing in existing scripts are still needed.

farukalamai:
Hi @Borda ! I've opened a PR for the rfdetr/evaluation and rfdetr/utilities modules: #888

hinogi:
how strict should the typing be? I experienced that pylance and such will 'wrongly' narrow types down to Unknown although they have been in instance check before.
Do you want an explicit cast in that case so the type checker is happy?

hinogi:
Well, I am not a python dev, but I know my way around stuff. There are a lot of open traps being laid out in the source code for unexperienced developers I can tell.
It is pretty messy.

Jonathan-Jesni:
Hi @Borda 
[...]

PR #1283, opened 2026-08-05
fix: guard `ModelContext.reinitialize_detection_head` against cleared weights

atikulmunna:
## Description

Found while working through #564. Standalone fix, no typing config changes here.

`ModelContext.model` is declared `torch.nn.Module`, and it is neither of those things.

**It is an `LWDETR`.** `inference.py` asserts `isinstance(nn_model, LWDETR)` immediately before the only construction site, and the class docstring already called it an LWDETR instance. Only the declaration stayed loose.

**It is also optional.** `RFDETR.inference(inplace=True)` sets it to `None` to free the weights, and three methods already raise `RuntimeError` when they find it `None`:

- `evaluate()`, detr.py:1062: `"Cannot evaluate: the base model has been cleared by a previous inplace optimization."`
- `export()`, detr.py:1571
- `deploy_to_roboflow()`, detr.py:2540

`reinitialize_detection_head()` is the one that reaches through `self.model` without that check, so calling it after an inplace optimization fails with:

```
AttributeError: 'NoneType' object has no attribute 'reinitialize_detection_head'
```

It now raises `RuntimeError` in the same style as the three existing checks. The guard runs before `args.num_classes` is assigned, so a rejected call cannot leave the context half updated.

Declaring the attribute `LWDETR | None` makes the type say what the code has always done.

## Changes

- `src/rfdetr/inference.py`: declare `ModelContext.model` as `LWDETR | None`, narrow the constructor parameter to `LWDETR`, guard `reinitialize_detection_head`, update the docstring
- `tests/inference/test_model_inference.py`: cover the guard

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

Narrowing the constructor parameter from `nn.Module` to `LWDETR` is a type-level tightening only. It matches the `isinstance` assert that already runs on the sole construction path, and nothing in the package passes anything else.

## Testing

`ModelContext.reinitialize_detection_head` had no coverage before this PR. It has no in-package callers, and every existing test matching that method name targets `LWDETR.reinitialize_detection_head` or `RFDETRModule.reinitialize_detection_head` instead, which are different methods.

I checked the two new tests are not vacuous: with the guard removed, both fail on the original `AttributeError`, which is the failure the guard replaces.

- `python -m mypy src/rfdetr/ --no-error-summary` exits 0
- `pre-commit run --all-files` passes, all 19 hooks
- Full CPU suite matching `ci-tests-cpu.yml`: 3301 passed, 59 skipped, 0 failed

## Notes

This also unblocks `rfdetr.detr` for #564. Three of its five strict errors come from this declaration, and with it corrected the rest is a one-line re-export plus two `pyproject.toml` lines. I kept that in a separate PR so this one can be judged as the fix it is. Happy to open it once this lands.

### DIFF
--- a/src/rfdetr/inference.py
+++ b/src/rfdetr/inference.py
@@ -65,11 +65,12 @@ class ModelContext:
             num_classes: New number of output classes (including background).
 
         Raises:
-            RuntimeError: If the model weights were already cleared by ``inference(inplace=True)``.
+            RuntimeError: If the model weights were already cleared by ``RFDETR.inference(inplace=True)``.
         """
         if self.model is None:
             raise RuntimeError(
-                "Cannot reinitialize the detection head after inference(inplace=True) cleared the model weights."
+                "Cannot reinitialize the detection head after inplace optimization. "
+                "The original model has been cleared. Create a new RFDETR instance."
             )
         reinitialize_head = cast("Callable[[int], None]", self.model.reinitialize_detection_head)
         reinitialize_head(num_classes)
--- a/src/rfdetr/inference.py
+++ b/src/rfdetr/inference.py
@@ -31,7 +31,9 @@ class ModelContext:
     ``populate_args()`` or the legacy stack.
 
     Args:
-        model: The underlying ``nn.Module`` (LWDETR instance).
+        model: The underlying ``LWDETR`` module. The attribute is cleared to ``None`` by
+            :meth:`RFDETR.inference` when called with ``inplace=True``, which frees the
+            weights from memory.
         postprocess: PostProcess instance for converting raw outputs to boxes.
         device: Device the model lives on.
         resolution: Input resolution (square side length in pixels).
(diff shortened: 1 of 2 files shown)


atikulmunna:
Both review comments addressed in 9117142.

**Docstring:** agreed, `ModelContext` has no `inference()` of its own so the bare reference was ambiguous. Now reads `RFDETR.inference(inplace=True)`, matching how the class docstring already refers to it.

**Error message:** agreed on the substance, the message gave no recovery path. It now follows the same three-part shape `RFDETR.export()` uses for this exact condition:

```
Cannot reinitialize the detection head after inplace optimization.
The original model has been cleared. Create a new RFDETR instance.
```

One correction on the premise, since it affects which message to copy. The three existing guards are not consistent with each other:

- `RFDETR.export()`, detr.py:1573: plain sentences, no dash, has a recovery step
- `RFDETR.evaluate()`, detr.py:1062: colon form, no dash, no recovery step
- `deploy_to_roboflow()`, detr.py:2542: dash form, has a recovery step

So only one of the three uses a dash, and `export()` is the closest analogue here (same "the weights are gone, start over" situation, whereas `deploy_to_roboflow()` can point at an alternative artifact). I matched `export()` rather than introduce a fourth variant. Happy to align all four in a follow-up if you want one house style, but that felt out of scope for this PR.

Also updated the test to assert on the leading clause rather than the wording that changed, so it is not coupled to the exact phrasing. Re-verified it still fails with the original `AttributeError` when the guard is removed.

Borda (approved):

PR #1288, opened 2026-08-05
refactor: enable mypy strict checking for rfdetr.datasets.yolo

atikulmunna:
## Description

Part of #564, following your go-ahead in https://github.com/roboflow/rf-detr/issues/564#issuecomment-5175564410.

Enables `mypy --strict` for `rfdetr.datasets.yolo` and removes it from the `ignore_errors` list. 23 errors to zero.

### The `ndarray` convention

Sixteen of the 23 were the bare-`ndarray` `type-arg` case I raised on the issue. You said go ahead without naming a convention, so I used the precise `NDArray[dtype]` form I had argued for. Flagging it clearly here because this is the PR that sets the precedent for `transforms`, `synthetic` and `save_grids` (10 more of the same error), and it is easy to change now and annoying to change later.

The reason it is the right call specifically in this file: the dtype is fixed at every construction site, so nothing is being guessed.

| what | dtype | where it is set |
|---|---|---|
| boxes, polygons, keypoints | `np.float32` | `_parse_yolo_box`, `_parse_yolo_polygon`, `_parse_yolo_pose_label_line` |
| `class_id` | `np.int64` | `np.array(class_id, dtype=np.int64)` |
| rasterized masks | `np.bool_` | `np.array(mask, dtype=bool)` |
| decoded images | `np.uint8` | `np.array(image.convert("RGB"))` |

If you would rather have `np.ndarray[Any, Any]` here to match `evaluation/`, say so and I will switch this PR before it lands.

### The other seven

- `UnidentifiedImageError` now comes from `PIL` directly. Reaching it via `Image.UnidentifiedImageError` is an implicit re-export that strict mode rejects.
- `ConvertYolo.__call__` declares its `dict` and `tuple` parameters, and `target_out` is declared `dict[str, Tensor]` to match what it is built from.
- `YoloDetection.__init__` declares `transforms` as `Compose | None`, and `__getitem__` declares its return type. The prepared image and target are bound to their own names so the pre-transform and post-transform types stay distinct.
- `keypoint_schema` is declared `YoloKeypointSchema | None` ahead of the branch that assigns it.
- `YoloDetection` subclasses `VisionDataset`, which is `Any`. Suppressed at the class with a note. I checked before suppressing: the installed torchvision ships no `py.typed`, so the existing `ignore_missing_imports` entry is correct and there is nothing to fix on our side.

### One behavioral detail worth your eyes

The dataset-selection branch now tests `self.keypoint_schema is not None` instead of `include_keypoints`:

```python
if self.keypoint_schema is not None:
    self.sv_dataset = _build_lazy_yolo_keypoint_dataset(...)
```

`_build_lazy_yolo_keypoint_dataset` requires a non-None schema, and mypy cannot carry the "`include_keypoints` implies the schema exists" invariant across the two separate blocks. The two conditions are equivalent: the schema is set exactly on the `include_keypoints` path, and a failed inference re-raises rather than leaving it `None`. The line just above already tests the schema this way, so this matches the file's existing idiom, and it makes the argument provable without an `assert`.

Since a wrong equivalence here would silently route pose datasets down the detection path, I ran the keypoint tests specifically rather than relying on the suite as a whole: `pytest tests/datasets/test_yolo.py -k "keypoint or pose"`, 16 passed.

## Changes

- `src/rfdetr/datasets/yolo.py`: type annotations as above
- `pyproject.toml`: drop `rfdetr.datasets.yolo` from the `ignore_errors` block

## Type of change

- [x] Refactor / typing (no functional change)

## Testing

Verified in an environment matching `ci-typing.yml` (Python 3.10, `-e ".[train,cli,visual]" --group typing`):

- `python -m mypy src/rfdetr/ --no-error-summary` exits 0
- `pre-commit run --all-files` passes, all 19 hooks
- `tests/datasets/`: 578 passed
- `tests/datasets/test_yolo.py -k "keypoint or pose"`: 16 passed
- Full CPU suite matching `ci-tests-cpu.yml`: 3335 passed, 59 skipped, 0 failed

## Notes

This leaves 7 modules in the block. Whichever convention you confirm above, I am happy to keep going through `transforms`, `synthetic` and `save_grids`, which share the same `ndarray` pattern.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -413,7 +413,6 @@ overrides = [
     "rfdetr.datasets.save_grids",
     "rfdetr.datasets.synthetic",
     "rfdetr.datasets.transforms",
-    "rfdetr.datasets.yolo",
     "rfdetr.detr",
   ], ignore_errors = true },
 ]
--- a/src/rfdetr/datasets/yolo.py
+++ b/src/rfdetr/datasets/yolo.py
@@ -17,7 +17,8 @@ import yaml
 
 if TYPE_CHECKING:
     from supervision import Detections
-from PIL import Image, ImageDraw
+from numpy.typing import NDArray
+from PIL import Image, ImageDraw, UnidentifiedImageError
 from torchvision.datasets import VisionDataset
 
 from rfdetr.datasets._aug_utils import resolve_keypoint_flip_pairs
@@ -27,6 +28,7 @@ from rfdetr.datasets._keypoint_schema import (
     _load_yaml_mapping,
     infer_yolo_keypoint_schema,
 )
+from rfdetr.datasets._torchvision import Compose
 from rfdetr.datasets.coco import (
     make_coco_transforms,
     make_coco_transforms_square_div_64,
@@ -42,7 +44,7 @@ REQUIRED_DATA_SUBDIRS = ["images", "labels"]
 YOLO_IMAGE_EXTENSIONS = {".bmp", ".dng", ".jpg", ".jpeg", ".mpo", ".png", ".tif", ".tiff", ".webp"}
 
 
-def _parse_yolo_box(values: list[str]) -> np.ndarray:
+def _parse_yolo_box(values: list[str]) -> NDArray[np.float32]:
     """Parse a YOLO center-width-height box into relative XYXY coordinates."""
     x_center, y_center, width, height = values
     return np.array(
@@ -56,7 +58,7 @@ def _parse_yolo_box(values: list[str]) -> np.ndarray:
     )
 
 
-def _box_to_polygon(box: np.ndarray) -> np.ndarray:
+def _box_to_polygon(box: NDArray[np.float32]) -> NDArray[np.float32]:
     """Convert a relative XYXY box into a 4-corner polygon."""
     return np.array(
         [[box[0], box[1]], [box[2], box[1]], [box[2], box[3]], [box[0], box[3]]],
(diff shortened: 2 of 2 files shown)


atikulmunna:
The GPU job is red on one test:

```
FAILED tests/benchmarks/test_training_synthetic.py::test_train_convergence_segmentation
AssertionError: val segm mAP 0.043 should reach at least 0.05 after Trainer.fit.
assert 0.04284944385290146 >= 0.05
```

Three reasons I do not think it relates to this PR.

The test already declares itself non-deterministic:

```python
@pytest.mark.gpu
@pytest.mark.flaky(reruns=1, only_rerun="AssertionError")
def test_train_convergence_segmentation(
```

`only_rerun="AssertionError"` is this exact failure class, and the `1 rerun` in the job log is that retry being spent before the job went red. It trains `RFDETRSegNanoConfig` from random init for 5 epochs at `batch_size=4` against a 5 percent floor, so some spread is expected.

No gpu-marked test references yolo. Grepping every file containing `pytest.mark.gpu` for a yolo reference returns nothing, so this workflow does not exercise the code this PR changes in either direction. The coverage that applies here is the CPU suite and the other 109 checks, all green.

It also passes locally on `5464a0a`, this exact commit, across two runs:

```
tests/benchmarks/test_training_synthetic.py::test_train_convergence_segmentation
1 passed in 228.83s
```

On a full local gpu suite run, `test_train_convergence_native_ptl` missed its floor instead. That is the other test carrying the same flaky marker.

Re-run it if you want the square green, but I do not think this job can tell you much about this PR either way.

Borda (approved):

PR #1303, opened 2026-08-06
fix(coco): drop unannotated grouping categories

Borda:
Roboflow COCO exports prepend a synthetic root category that every real class lists as its supercategory and that owns no annotations. CocoDetection built cat2label from every entry of coco.cats, so that root took label index 0 and an extra output channel, and _detect_num_classes_for_training counted the same unfiltered list.

Add filter_parent_categories() and annotated_category_ids() to rfdetr.datasets.coco. A category is dropped only when it is named as another category's supercategory and no annotation references its id, so annotated parents keep their label slot and ConvertCoco does not raise KeyError on them. Flat datasets and pathological all-parent inputs are returned unchanged.

CocoDetection.cat2label, RFDETR._detect_num_classes_for_training and RFDETR._load_classes now share that filter, the duplicated hierarchy logic in _load_classes is removed, and skipped category ids are logged. Training a hierarchical export builds an N-class head instead of N+1 and shifts real class label indices down one; the keypoint path _build_keypoint_cat2label is untouched.

Rewrite test_coco_auto_detect_uses_full_category_mapping_not_leaf_only_names, which asserted the unfiltered count, into separate unannotated-grouping and annotated-parent cases, and add regression tests covering cat2label, label2cat, target labels, class-name alignment and the count-equals-remap invariant.


src/rfdetr/datasets/coco.py
```diff
@@ -45,6 +46,22 @@
...
+    "filter_parent_categories",
+    "compute_multi_scale_scales",
+    "convert_coco_poly_to_mask",
+    "CocoDetection",
+    "ConvertCoco",
+    "make_coco_transforms",
+    "make_coco_transforms_square_div_64",
+    "build_coco",
+    "build_roboflow_from_coco",
+]
```
Borda:
```suggestion
```

PR #1308, opened 2026-08-07
fix(datasets): serialize concurrent downloads of the same COCO asset

JESUSROYETH:
## What does this PR do?

While chasing an `EOFError` on `Testing (GPU tests Workflow)` in #1297, I traced it to a race in the benchmark download fixtures, not to the diff of that PR or to a bad transfer from `images.cocodataset.org`.

`tests/benchmarks/conftest.py` has three session-scoped fixtures that can each trigger `_download_and_extract(_COCO_URLS["annotations"], _DATA_DIR)` for the exact same zip: `download_coco_val`, `download_coco_val_keypoints`, `download_coco_train_val_keypoints`. Each guards its own check-then-download sequence with its own lock file (`.coco_download.lock`, `.coco_keypoint_download.lock`, `.coco_keypoint_train_val_download.lock`). Under `pytest -n 3`, a session-scoped fixture runs once per xdist worker, not once globally, so if two of these fixtures are needed at the same time on two different workers, both see "annotations file missing" and both call `_download_and_extract` at once — each holding a different lock, so neither blocks the other. One worker's `urlretrieve` can then start overwriting `annotations_trainval2017.zip` on disk while the other is mid-read inside `zipfile`, which surfaces as a bare `EOFError` (`zipfile/__init__.py:1108`, inside `_read2`).

This is why it's intermittent: #1306 (merged just before my last rebase) added retries for `BadZipFile`/`zlib.error`/`OSError`, which covers a genuinely truncated network transfer, but this failure mode isn't a bad transfer — the file on disk is fine until a second process starts rewriting it mid-read. `EOFError` doesn't subclass any of the three, so the retry never fired. Same commit on `develop`, `fix/dwn`, and my branch: some runs the two fixtures happen to not overlap and it's green, some runs they do and it's red.

## Changes

- `_download_and_extract` now takes a lock keyed by the destination zip filename (derived from the URL, same as `zip_path`) around the whole download+extract sequence, so any two callers requesting the same asset are serialized regardless of which fixture called them or which lock file that fixture happened to pick.
- Added `EOFError` to the retried exceptions, since a torn read of a zip that's being concurrently rewritten is a legitimate transient failure the retry loop should absorb, same as `BadZipFile`/`zlib.error`.

I left the three fixture-level locks in `conftest.py` as-is — they're harmless now, and narrowing this PR to `_develop.py` keeps the fix in the one place that actually owns the shared resource.

## Validation

- New regression test (`test_concurrent_callers_for_the_same_url_do_not_race`) drives two threads through `_download_and_extract` for the same URL/`dest_dir`, pausing the first inside `_extract_zip` and launching the second while it's paused. Asserts the second caller's fetch never starts until the first has fully finished.
- Confirmed the test discriminates: reverted just the lock change and reran it — fails (`assert 3 == 2`, and the captured log shows a real `[Errno 2] No such file or directory` from the two calls stepping on the same `zip_path`, the same failure class as the original `EOFError`). Re-applied the fix, reran 8 times in a row: 12/12 pass each time, no flake.
- `pytest tests/benchmarks/test_develop_downloads.py -q`: 12 passed.
- `pytest --doctest-plus src/rfdetr/datasets/_develop.py -q`: 6 passed (docstring examples still valid after the edit).
- `ruff check` / `ruff format --diff` on both changed files: clean.
- `mypy src/rfdetr/datasets/_develop.py`: no issues.

I didn't re-run the GPU workflow's full COCO-download tests against the real network for this PR — the regression test reproduces the exact interleaving locally without needing the 250 MB download, and the fix is a lock-scoping change, not a change to the network path itself.

### DIFF
--- a/src/rfdetr/datasets/_develop.py
+++ b/src/rfdetr/datasets/_develop.py
@@ -238,9 +238,19 @@ def _extract_zip(zip_path: Path, dest_dir_resolved: Path) -> None:
 def _download_and_extract(url: str, dest_dir: Path) -> None:
     """Download a zip archive and safely extract it, retrying on a corrupt transfer.
 
-    Truncated downloads and corrupt deflate streams (``zlib.error``/``BadZipFile``) are
-    retried up to ``_DOWNLOAD_MAX_ATTEMPTS`` times. The partial archive is always removed,
-    both on failure (so a corrupt file never persists for the next run) and on success.
+    Truncated downloads and corrupt deflate streams (``zlib.error``/``BadZipFile``/
+    ``EOFError``) are retried up to ``_DOWNLOAD_MAX_ATTEMPTS`` times. The partial archive
+    is always removed, both on failure (so a corrupt file never persists for the next run)
+    and on success.
+
+    Concurrent calls for the same *url* are serialized on a lock keyed by the destination
+    zip filename, derived the same way as ``zip_path`` below. This matters because several
+    independent callers (for example the ``download_coco_val`` and
+    ``download_coco_val_keypoints`` pytest fixtures, each running in its own
+    pytest-xdist worker) can end up requesting the same archive at the same time. Without
+    a shared lock, one caller's ``urlretrieve`` can start overwriting the archive on disk
+    while another caller is mid-extraction, which is what surfaces as the ``EOFError``
+    below rather than anything ``_fetch_zip``'s size check catches.
 
     Args:
         url: URL to a zip archive.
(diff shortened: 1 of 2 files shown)


JESUSROYETH:
Closing — Borda pushed essentially the same fix directly to #1297's branch a few minutes ago (70cd542c), and it goes further than this PR: unique per-process staging filenames so two callers never share one archive path at all, plus an `is_complete` recheck after the lock to skip a redundant re-fetch. No need for two PRs carrying the same fix.

PR #1350, opened 2026-08-17
Read a scalar CLAHE `clip_limit` as a range, matching albumentations

### LINKED ISSUE #1349: CLAHE clip_limit: Kornia backend pins a scalar where Albumentations samples a range
### Search before asking

### Bug

On the Kornia GPU backend, `CLAHE` applies a fixed `clip_limit` where the CPU (Albumentations) backend samples one from a range. This affects the **default** configuration, so it is not limited to unusual configs.

Albumentations expands a scalar `clip_limit` to `(1, v)`:

```python
>>> import albumentations as A
>>> A.CLAHE(clip_limit=4.0).clip_limit
(1.0, 4.0)
```

`_make_clahe` in `src/rfdetr/datasets/kornia_transforms.py` routes it through `_as_range`, which expands a scalar to `(v, v)`:

```python
return random_clahe(
    clip_limit=_as_range(params.get("clip_limit", 4.0)),
    ...
)
```

`4.0` is the default on both sides, so with no user configuration at all the two backends disagree:

| aug_config | Kornia `clip_limit` | Albumentations `clip_limit` |
|---|---|---|
| `{}` (default) | `(4.0, 4.0)` | `(1.0, 4.0)` |
| `{"clip_limit": 4.0}` | `(4.0, 4.0)` | `(1.0, 4.0)` |
| `{"clip_limit": 2.0}` | `(2.0, 2.0)` | `(1.0, 2.0)` |
| `{"clip_limit": (1.0, 4.0)}` | `(1.0, 4.0)` | `(1.0, 4.0)` |

Only the explicit-pair form agrees. Every scalar form, including the default, pins the GPU path to maximum contrast enhancement on every sample while the CPU path varies it, and nothing is logged.

The docstring states the opposite:

> Both parameters map directly: Albumentations' ``clip_limit`` (a scalar or a pair) becomes Kornia's ``clip_limit`` range

`_as_range`'s own docstring is accurate about what it does — it expands a scalar to the degenerate `(v, v)` — so this is a wrong choice of helper at the call site rather than a bug inside the helper.

### Expected behavior

A scalar `clip_limit` should reach Kornia as `(1, v)`, matching how Albumentations reads the same config, so the two backends draw from the same distribution.

```python
import albumentations as A
from rfdetr.datasets.kornia_transforms import _make_clahe

print(_make_clahe({}).clip_limit)              # (4.0, 4.0)
print(A.CLAHE().clip_limit)                    # (1.0, 4.0)
```

### Environment

- RF-DETR: `develop` at `b18c9ca`
- kornia 0.8.3, albumentations 2.0.8, torch 2.6.0
- CPU only; no GPU or model weights needed to reproduce

### Additional

Introduced in #1277, which added the CLAHE mapping — mine, so I would like to fix it.

I checked the other factories that call `_as_range` for the same class of mismatch:

- `Sharpen` (`alpha`) and `GaussNoise` (`std_range`) — Albumentations rejects a scalar for both with `ValueError`, so there is no scalar form to disagree about.
- `GaussianBlur` (`sigma`) — Albumentations expands a scalar `sigma_limit` to `(0, v)` rather than `(v, v)`, so the same gap exists, but only when a user passes a scalar; the configured default is already a pair. Happy to include it here or leave it separate, whichever you prefer.

adhavan18:
Fixes #1349.

## The bug

Albumentations applies `to_tuple(clip_limit, low=1)`, so a scalar `v` means the **range** `(1, v)` and it samples from it per call. `_make_clahe` routed the value through `_as_range`, which expands a scalar to the degenerate `(v, v)`.

`4.0` is the default on both sides, so this applied with no user configuration at all:

| aug_config | before | after | albumentations |
|---|---|---|---|
| `{}` (default) | `(4.0, 4.0)` | `(1.0, 4.0)` | `(1.0, 4.0)` |
| `{"clip_limit": 4.0}` | `(4.0, 4.0)` | `(1.0, 4.0)` | `(1.0, 4.0)` |
| `{"clip_limit": 2.0}` | `(2.0, 2.0)` | `(1.0, 2.0)` | `(1.0, 2.0)` |
| `{"clip_limit": (2.0, 6.0)}` | `(2.0, 6.0)` | `(2.0, 6.0)` | `(2.0, 6.0)` |

Only the explicit-pair form agreed. Every scalar form, including the default, pinned the GPU path to maximum contrast enhancement on every sample while the CPU path varied it — a silent train/serve divergence with no warning, since from the code's point of view nothing was being collapsed.

This came in with #1277, which added the CLAHE mapping. Mine.

## The fix

A dedicated `_as_clahe_clip_limit` rather than a special case inside `_as_range`: the helper is correct about what it does and is used by transforms whose scalar semantics really are `(v, v)`. Changing it there would have fixed CLAHE and broken the others.

A pair is used as given, so only the scalar form changes.

## Scope

I checked the other `_as_range` call sites for the same class of mismatch:

- **`Sharpen`** (`alpha`) and **`GaussNoise`** (`std_range`) — albumentations rejects a scalar for both with `ValueError`, so there is no scalar form to disagree about. Safe.
- **`GaussianBlur`** (`sigma`) — albumentations expands a scalar `sigma_limit` to `(0, v)`, so the same gap exists, but only when a user passes a scalar; the configured default is already a pair. Left out of this PR deliberately — happy to fold it in here or do it separately, whichever you prefer.

## Verification

Nine assertions, all five configured forms plus direct parity against albumentations:

```
ok  param=None       -> (1.0, 4.0)
ok  param=4.0        -> (1.0, 4.0)
ok  param=2.0        -> (1.0, 2.0)
ok  param=(1.0, 4.0) -> (1.0, 4.0)
ok  param=(2.0, 6.0) -> (2.0, 6.0)
--- parity with albumentations ---
ok  clip_limit=4.0        kornia=(1.0, 4.0) albu=(1.0, 4.0)
ok  clip_limit=2.0        kornia=(1.0, 2.0) albu=(1.0, 2.0)
ok  clip_limit=(2.0, 6.0) kornia=(2.0, 6.0) albu=(2.0, 6.0)
```

The three scalar cases fail on the parent commit:

```
FAIL param=None  -> (4.0, 4.0)  expected (1.0, 4.0)
FAIL param=4.0   -> (4.0, 4.0)  expected (1.0, 4.0)
FAIL param=2.0   -> (2.0, 2.0)  expected (1.0, 2.0)
```

`test_clahe_maps_both_parameters` still passes — `grid_size` and the explicit pair are unaffected.

`ruff check` and `ruff format --check` clean on both files. `pytest` cannot collect the module in my environment (`transformers` is missing `BackboneConfigMixin`, unrelated to this diff), so I exercised the factory directly; CI here will be the better check.

### DIFF
--- a/src/rfdetr/datasets/aug_configs.py
+++ b/src/rfdetr/datasets/aug_configs.py
@@ -89,7 +89,7 @@ or torchvision defaults. Install it with ``pip install 'rfdetr[augment]'``.
 | ``Blur`` | ``K.RandomBoxBlur`` | Box blur; ``blur_limit`` rounded up to odd, pair collapses to its upper bound |
 | ``Sharpen`` | ``K.RandomSharpness`` | ``sharpness = 1.0 + alpha`` (1.0-pivoted); ``lightness``/``method`` ignored |
 | ``Equalize`` | ``K.RandomEqualize`` | Only ``p`` honored; ``mode``/``by_channels``/``mask`` ignored |
-| ``CLAHE`` | ``K.RandomClahe`` | ``clip_limit`` and ``tile_grid_size`` map directly |
+| ``CLAHE`` | ``K.RandomClahe`` | Scalar ``clip_limit=v`` -> ``(1, v)``; pairs and ``tile_grid_size`` pass through |
 | ``Perspective`` | ``K.RandomPerspective`` | Approximate; see the note below. ``keep_size=False`` raises |
 
 ``Perspective`` is the one entry in the table that is not a faithful mapping. Albumentations samples each
--- a/src/rfdetr/datasets/kornia_transforms.py
+++ b/src/rfdetr/datasets/kornia_transforms.py
@@ -594,18 +594,47 @@ def _make_equalize(params: dict[str, Any]) -> Any:
     return RandomEqualize(p=params.get("p", 0.5))
 
 
+def _as_clahe_clip_limit(value: Any) -> tuple[float, float]:
+    """Normalise ``CLAHE``'s ``clip_limit`` the way Albumentations does.
+
+    Albumentations applies ``to_tuple(clip_limit, low=1)``, so a scalar ``v`` means the range ``(1, v)``. This differs
+    from :func:`_as_range`, which expands a scalar to the degenerate ``(v, v)``; using that here would change the
+    distribution rather than the units.
+
+    Args:
+        value: A scalar, a 1-element sequence, or a 2-element ``(min, max)`` pair.
+
+    Returns:
+        The ``(min, max)`` pair Albumentations would sample from for the same config.
+
+    Examples:
+        >>> _as_clahe_clip_limit(4.0)
+        (1.0, 4.0)
+        >>> _as_clahe_clip_limit((2.0, 6.0))
+        (2.0, 6.0)
+    """
+    if isinstance(value, (list, tuple)) and len(value) == 2:
+        return (float(value[0]), float(value[1]))
+    low, high = _as_range(value)
+    # A scalar (or 1-element sequence) arrives here as the degenerate (v, v).
+    return (1.0, high) if low == high else (low, high)
+
+
 def _make_clahe(params: dict[str, Any]) -> Any:
     """Build a ``K.RandomClahe`` from aug_config ``CLAHE`` params.
 
-    Both parameters map directly: Albumentations' ``clip_limit`` (a scalar or a pair) becomes Kornia's ``clip_limit``
-    range, and ``tile_grid_size`` becomes ``grid_size``.
+    ``tile_grid_size`` becomes ``grid_size`` directly. ``clip_limit`` needs care: Albumentations reads a scalar ``v``
+    as the range ``(1, v)`` and samples from it, not as the fixed value ``v``. Passing it through :func:`_as_range`
+    would produce the degenerate ``(v, v)`` and pin the GPU path to maximum contrast enhancement on every sample while
+    the CPU path varied it — and since ``4.0`` is the default on both sides, that divergence applied to the default
+    configuration rather than only to unusual ones. A pair is already a range and is used as given.
     """
     import kornia.augmentation as kornia_augmentation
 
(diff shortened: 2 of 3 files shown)


adhavan18:
﻿all three addressed in `e01abf0`. the first one is a real behaviour bug and the reviewer is right about the mechanism.

## the one-element sequence

confirmed against the pinned albumentations before changing anything:

```
4.0            -> ACCEPTED, normalised to (1.0, 4.0)
(1.0, 4.0)     -> ACCEPTED, normalised to (1.0, 4.0)
[1.0, 4.0]     -> ACCEPTED, normalised to (1.0, 4.0)
[4.0]          -> REJECTED: ValueError: 2 validation errors for InitSchema
(4.0,)         -> REJECTED: ValueError: 2 validation errors for InitSchema
```

so `[4.0]` is refused on the CPU path, while this helper routed it through `_as_range`, got the degenerate `(4.0, 4.0)`, and turned it into `(1.0, 4.0)`. that is the same class of divergence the PR set out to remove, just pointing the other way: a config that trains on the GPU backend and raises on the CPU one.

`_as_clahe_clip_limit` now raises on any sequence that is not exactly two elements, and says why, naming the CPU backend so the message is actionable rather than just a type complaint. scalars are handled directly instead of borrowing `_as_range`, since the degenerate pair was only ever an intermediate step and it is what made the one-element case look scalar.

## the two test conventions

- annotated both parametrized arguments.
- split `test_clahe_clip_limit_matches_albumentations` into three independent cases with ids (`scalar-default-value`, `scalar`, `pair`), so a disagreement on one value no longer hides the other two behind a single reported failure.
- added `test_clahe_rejects_sequences_that_albumentations_rejects`, parametrized over `[4.0]`, `(4.0,)` and a three-element sequence.

`118 passed`, `ruff check` and `ruff format --check` clean.

adhavan18:
Bumping this since it's been quiet for a couple weeks — just rebased onto current develop, no conflicts. Happy to address anything if it needs another look.

Borda:
Hi, apologies for delay, I was off last week but I have on my radar for later this week...

Borda (approved):

PR #1359, opened 2026-08-17
refactor: enable mypy strict checking for rfdetr.datasets.synthetic

sahilmehta17:
Part of #564. Removes `rfdetr.datasets.synthetic` from the `ignore_errors` list in `pyproject.toml` and fixes the errors that surfaced.

### Typing

Most were the `type-arg` errors tracked in the issue: bare `np.ndarray` annotations on `draw_synthetic_shape`, `calculate_boundary_overlap`, and `generate_synthetic_sample`.

I used `numpy.typing.NDArray` with concrete dtypes rather than `np.ndarray[Any, Any]`, following `rfdetr.datasets.yolo`. The convention came up twice in #564 without an explicit ruling, so I matched the most recently merged module. Dtypes are read off the code rather than guessed: images are `uint8` (`np.ones(..., dtype=np.uint8)`), boxes are `float64` (`np.array` over Python floats). Happy to switch to `np.ndarray[Any, Any]` if you would rather keep the older form here.

`generate_synthetic_sample`'s local `img` also needed an explicit annotation. `np.ones` with a literal 3-tuple infers `ndarray[tuple[int, int, int], ...]`, which will not accept the shape-generic array `draw_synthetic_shape` returns, so the reassignment inside the placement loop failed. Declaring `img` as `NDArray[np.uint8]` keeps the variable shape-generic.

The remaining two were a missing return annotation on `DatasetSplitRatios.__post_init__` and an element type for `xyxys`.

### One bug, surfaced by the annotations

Separate logical change, called out so it is easy to review or ask me to split.

`_write_coco_json` indexed `detections.class_id` without checking it:

```python
class_id = int(detections.class_id[det_idx])
```

`Detections.class_id` is `Optional` in supervision, and `_write_coco_json` takes an arbitrary `detections_list` from its caller. Detections built without class ids therefore raised `TypeError: 'NoneType' object is not subscriptable` at the subscript.

It now raises `ValueError` naming the offending image, matching how the function already reports mismatched list lengths and missing polygon entries. The docstring `Raises:` section is updated.

`test_write_coco_json_raises_when_class_id_missing` covers it. It fails on the unfixed module with the `TypeError` above and passes with the guard.

### Verification

Rebased on `develop` at 92189f6.

- `mypy src/rfdetr/ --no-error-summary`: clean, run against the versions the typing CI resolves (`mypy==1.19.1`, `numpy==2.2.6`, `supervision==0.30.0`)
- `pre-commit run --all-files`: all hooks pass
- `pytest tests/datasets/`: 400 passed against a 399-passed baseline on `develop`, same skips, same pre-existing local failures on both sides. The delta is the one new test

No runtime behaviour changes beyond the new guard.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -408,7 +408,6 @@ overrides = [
   { module = [
     "rfdetr.config",
     "rfdetr.datasets.coco",
-    "rfdetr.datasets.synthetic",
     "rfdetr.datasets.transforms",
   ], ignore_errors = true },
 ]
--- a/src/rfdetr/datasets/synthetic.py
+++ b/src/rfdetr/datasets/synthetic.py
@@ -18,6 +18,7 @@ from pathlib import Path
 from typing import Literal
 
 import numpy as np
+from numpy.typing import NDArray
 from PIL import Image
 from supervision import Color, Detections, box_iou_batch, draw_filled_polygon
 from tqdm.auto import tqdm
@@ -42,7 +43,7 @@ class DatasetSplitRatios:
     val: float = 0.2
     test: float = 0.1
 
-    def __post_init__(self):
+    def __post_init__(self) -> None:
         """Validate that ratios sum to approximately 1.0 and are non-negative."""
         total = self.train + self.val + self.test
         if any(r < 0 for r in [self.train, self.val, self.test]):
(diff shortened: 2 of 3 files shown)


Borda (dismissed):

Borda:
> Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/roboflow/rf-detr?pullRequest=1359) before we can accept your contribution.

@sahilmehta17 ^^ or share a print screen 🦝

sahilmehta17:
﻿Pushed a fix for the Type Check failure and re-ran the gate.

`generate_synthetic_sample`'s local `img` was inferred as `ndarray[tuple[int, int, int], dtype[uint8]]` from `np.ones` with a literal 3-tuple, so reassigning it from `draw_synthetic_shape` (shape-generic `NDArray[np.uint8]`) failed the assignment check. It is now declared as `NDArray[np.uint8]`.

Worth noting for anyone else working through #564: this only reproduces on the numpy the typing CI resolves. I had numpy 2.3.5 locally and `mypy src/rfdetr/` was clean; pinning to CI's `numpy==2.2.6` reproduced it exactly. The two `no-any-return` errors I was seeing in `evaluation/coco_eval.py` were the same artifact in reverse and are also absent at 2.2.6.

`mypy src/rfdetr/` and `pre-commit run --all-files` are both clean now against the CI-resolved versions.

Borda (approved):

PR #1364, opened 2026-08-18
Fix incorrect TFLite INT8 calibration guidance

### LINKED ISSUE #1363: TFLite INT8 docs and warning incorrectly require calibration data
### Search before asking

- [x] I searched the RF-DETR issues and found no similar bug report. 

#1024 discusses export formats broadly but does not track this problem.

### Bug

RF-DETR’s TFLite documentation says INT8 export requires representative calibration images for good accuracy. Export also warns that omitting `calibration_data` will produce poor quantization accuracy.

RF-DETR does not use calibration for INT8 export. `quantization="int8"` produces a dynamic-range model with INT8 weights and floating-point activations.

**Actual behavior**:

```python
model.export(
    format="tflite",
    quantization="int8",
    calibration_data=None,
)
```

Export succeeds and produces a `*_dynamic_range_quant.tflite` artifact, but logs:

```text
No calibration_data provided for INT8 quantization. Using random data — this will produce poor quantization accuracy.
```

That warning is incorrect. `_quantize_dynamic_range()` sets only:

```python
converter.optimizations = [tf.lite.Optimize.DEFAULT]
```

It never sets `converter.representative_dataset`. RF-DETR also never passes `output_integer_quantized_tflite` to `onnx2tf`, so calibration data cannot affect the generated INT8 model.

I also checked every currently supported `onnx2tf` release (2.4.0–2.4.3). None defines the `download_test_image_data()` function that `_patch_validation_download()` tries to patch, so supplied `calibration_data` does not currently reach `onnx2tf` validation either.

**Expected behavior**:

- Do not warn when `calibration_data` is omitted.
- Document `"int8"` as dynamic-range quantization that needs no calibration data.
- Describe `calibration_data` and `max_images` according to their current behavior.
- Keep static/full-integer INT8 unsupported.

### Environment

- RF-DETR 1.9.3
- `develop` at `50f21220`
- Python 3.12
- `onnx2tf` 2.4.3

```python
from rfdetr import RFDETRSmall

model = RFDETRSmall()
model.export(
    format="tflite",
    quantization="int8",
    calibration_data=None,
)
```

amanharshx:
Fixes #1363.

## Summary

TFLite `quantization="int8"` uses dynamic-range quantization: weights are quantized to INT8, activations remain floating point, and calibration data does not affect the generated models.

This PR removes the incorrect missing-calibration accuracy warning and updates the export documentation and docstrings. Public arguments, data preparation, quantization behavior, and static INT8 support remain unchanged.

## Validation

- TFLite export tests: 79 passed
- Export test suite: 375 passed, 67 skipped
- Ruff, formatting, codespell, and mypy on touched files pass


Borda:
@amanharshx, I see mostly just doc changes, no code or dependency updates, so you say it was working all the time? I think I tried it and it did not work for me... 😕 
btw, we know it shall work in theory, but in practice we hit some problems 🦝

amanharshx:
@Borda No, I’m not saying TFLite INT8 always worked end to end. This PR only fixes the calibration guidance. `_quantize_dynamic_range()` uses `Optimize.DEFAULT` without a `representative_dataset`, so `calibration_data` can’t change the generated model or fix an export failure.

I’ve hit separate export failures too. #1323 fixed the import-order deadlock and shipped in 1.9.3. I also reproduced a 1.9.3 PATH issue on macOS arm64 with Python 3.12, TensorFlow 2.19.1, and onnx2tf 2.4.3: invoking `<venv>/bin/python` without `<venv>/bin` on `PATH` prevents onnx2tf from launching `onnxsim`, and conversion later fails with the Functional-model error. Adding `<venv>/bin` to `PATH` makes the same export produce FP32 and FP16 models.

I have a fix and will file it separately with the full reproduction and logs. If what you hit was different, I’d be happy to have a look 🙂

Borda:
So let's add a cookbook with TFlite to validate all is fine now

Borda (approved):

PR #1370, opened 2026-08-19
feat(augment): support `ShiftScaleRotate` on the Kornia backend

adhavan18:
## What

Adds `ShiftScaleRotate` to the Kornia GPU augmentation backend. Follow-up to #1330 and #1277, same issue (#1252).

## Why

`ShiftScaleRotate` is one of the names `aug_configs` documents and `transforms.py` accepts on the CPU path, while the Kornia registry rejects it. Backend selection is automatic, so the same `aug_config` trains on a CPU box and raises on a GPU box:

```text
kornia ShiftScaleRotate -> ValueError: Unknown augmentation key 'ShiftScaleRotate' for Kornia GPU backend
albu   ShiftScaleRotate -> OK
```

It is safe to add for the same reason `Perspective` was: **it preserves the output resolution.** The GPU path rebuilds each batch as `NestedTensor(img_aug, samples.mask)`, reusing the pre-augmentation padding mask, so a transform that resizes leaves the mask describing a different shape. That is why the crops stay unsupported, and it does not apply here.

I should correct something I wrote in #1330 while I am here. The doc block excluded this name along with the crops, on the grounds that the group "move boxes and masks". That reason does not hold for this one: `Affine` moves boxes too and has always been supported. Resolution change is the real constraint, and `ShiftScaleRotate` does not change resolution.

## The part that is not a pass-through

Albumentations calls this "a special case of Affine transform", so an alias to `_make_affine` looks obvious. It would be wrong, because the limits are **deltas rather than absolute ranges**.

`scale_limit` is biased by 1 — from the Albumentations docstring:

> Note that the scale_limit will be biased by 1. If scale_limit is a tuple, like (low, high), sampling will be done from the range (1 + low, 1 + high).

So the documented default `(-0.1, 0.1)` means a scale between `0.9` and `1.1`. Kornia's `scale` is the absolute multiplier, so forwarding the raw value would ask it to scale the image to between a tenth of its size and nothing at all. Confirmed against the installed Albumentations rather than trusting the docstring:

```python
>>> A.ShiftScaleRotate(scale_limit=0.1).scale_limit
(0.9, 1.1)
```

The pivot is applied here, the same way `_make_sharpen` already pivots `alpha` at `1.0`.

Two smaller differences in the same vein:

| param | Albumentations | mapped as |
|---|---|---|
| `scale_limit` | delta biased by 1 | `scale=(1 + low, 1 + high)` |
| `shift_limit` | signed fraction range | `translate`, Kornia's non-negative per-axis maximum, as `_make_affine` already does for `translate_percent` |
| `shift_limit_x` / `shift_limit_y` | per-axis override | Kornia's `translate=(tx, ty)` expresses this directly |
| `rotate_limit` | degrees | `degrees` |

All three limits also read a scalar `v` as the symmetric `(-v, v)`, not the degenerate `(v, v)` that `_as_range` gives, so they go through a small `_as_symmetric_range` helper instead.

`interpolation`, `border_mode`, `mask_interpolation`, `fill`, `fill_mask` and `rotate_method` have no `RandomAffine` equivalent and are ignored **with a warning**, matching how `ToGray` and `Sharpen` already report their dropped parameters.

## One thing worth your call

**Albumentations deprecates this name.** Constructing it emits:

```
UserWarning: ShiftScaleRotate is a special case of Affine transform. Please use Affine transform instead.
```

I mapped it anyway because the CPU path in this repo still accepts it, so the parity gap is real for any config already using it, and `albumentations<3.0.0` is pinned. But if you would rather close this name out of `aug_configs` and point users at `Affine` — which this backend already supports — that is a reasonable call and I am happy to send that PR instead. The docs note added here says new configs should prefer `Affine`.

## Tests

14 cases in `TestShiftScaleRotateFactory`, covering the delta semantics (scalar and asymmetric pair), symmetric scalar expansion, the Albumentations defaults, per-axis shifts, each ignored option warning, no spurious warning on an ordinary config, resolution preservation, and boxes moving with the image.

Reading the resolved ranges needs `RandomAffine._param_generator` — `flags` carries only the resampling options — so there is a helper that fails with an actionable message if a future Kornia release changes that shape, mirroring the existing `_sharpness_sampler_range`.

```
121 passed
```

### DIFF
--- a/src/rfdetr/datasets/aug_configs.py
+++ b/src/rfdetr/datasets/aug_configs.py
@@ -108,8 +108,11 @@ apply. Its limits are deltas rather than absolute ranges, so they are converted 
 ``scale_limit`` is biased by 1 in Albumentations, meaning it samples from ``(1 + low, 1 + high)``, so the
 default ``(-0.1, 0.1)`` is a scale between ``0.9`` and ``1.1``; that pivot is applied here. All three limits
 also read a scalar ``v`` as the symmetric ``(-v, v)``. ``shift_limit_x`` and ``shift_limit_y`` map onto
-Kornia's per-axis ``translate``. ``interpolation``, ``border_mode``, ``mask_interpolation``, ``fill``,
-``fill_mask`` and ``rotate_method`` have no equivalent and are ignored with a warning. Note that
+Kornia's per-axis ``translate``; ``None`` uses the shared ``shift_limit``. Kornia can sample only symmetric shift
+ranges, so an asymmetric Albumentations pair such as ``(0.1, 0.2)`` becomes ``(-0.2, 0.2)`` and emits a warning.
+Use the Albumentations backend when that distribution must be preserved. ``interpolation``, ``border_mode``,
+``mask_interpolation``, ``fill``, ``fill_mask`` and ``rotate_method`` have no equivalent and are ignored with a warning.
+Note that
 Albumentations itself deprecates ``ShiftScaleRotate`` in favour of ``Affine``, which this backend already
 supports; new configs should prefer ``Affine``.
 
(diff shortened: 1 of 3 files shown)


adhavan18:
Bumping this since it's been quiet for a couple weeks — just rebased onto current develop, CI green across the full matrix, no conflicts. Happy to address anything if it needs another look.

Borda:
Hi, apologies for delay, I was off last week but I have on my radar for later this week...

Borda (approved):

PR #1374, opened 2026-08-19
perf(postprocess): preallocate mask output instead of concatenating chunks

JESUSROYETH:
`PostProcess._postprocess_masks` (right after #1369) resizes the selected K mask logits in chunks of `_MASK_CHUNK=32`, thresholds each chunk to bool, and appends it to a list. Then `torch.cat(chunks, dim=0)` builds the final `[K,1,H,W]` output. That last step duplicates the whole thing: every bool chunk stays alive in the list while `torch.cat` allocates a second same-size tensor and copies them all in. At K=300 masks resized to 1920x1080 the bool output alone is ~622 MB, so for a moment you hold close to the double of that.

**Changes**

- Writes each thresholded chunk straight into a preallocated destination (`masks_i.new_empty(...)`, then `torch.gt(interpolated, 0.0, out=...)`) instead of building the list + `torch.cat`. Bit-identical output; the write path is no longer gated to a single device (verified on CPU and CUDA).
- `test_postprocess.py`: one test pinning this on CPU (asserts `torch.cat` never runs, output matches the reference chunked computation), one on CUDA doing the same, and one CPU-only regression test asserting each chunk's buffer is actually dropped (via `weakref`) before the next chunk is built, so a future removal of the `del interpolated` fix below would fail it.

**A bug in my own first version of this**

My first draft only applied this on CPU. On CUDA a real forward pass (`RFDETRSegNano`, real image, K=100 masks) showed the new path using 17% *more* peak memory than the old one, so I kept CUDA on the original code and left it there.

Going back to check why, with `torch.cuda.memory_stats()` and a step-by-step allocation trace, I found it: `interpolated = F.interpolate(...)` inside a plain `for` loop reassigns the same variable each pass. Python builds the new value before dropping the old one, so for a moment the *previous* chunk's transient buffer and the *new* one are both alive — the same problem this PR fixes, just smaller and inside my own loop this time.

One `del interpolated` right after it's consumed fixes it: the CUDA regression on the real model drops from 17% to 3.75%, and a large synthetic case that was already a modest win becomes a much bigger one (peak memory down ~29% instead of ~7%). Output stays bit-identical everywhere I checked.

A small residual (2-4%) CUDA memory increase is still there in the 19-200 MB output range. I chased it further instead of leaving it unexplained, it comes down to chunk count. This path's CUDA peak is always `full output + at most one transient chunk`, held flat for the whole loop. The old path's peak is the larger of a mid-loop moment (`fraction done so far + one transient chunk`) or the final `torch.cat`'s full doubling. With few chunks (`num_select` at or below 4x `_MASK_CHUNK` = 128, so 4 chunks or fewer) the mid-loop moment stays just under this path's flat peak. With more chunks the cat's doubling dominates instead, and this path is clearly better. Confirmed with a sweep holding output size fixed while varying chunk count (1 chunk: 0% difference; 10 chunks at the same size as the regressing case: -23%, not a regression), then against every `num_select` a shipped segmentation config actually uses: 100 (`Nano`/`Small`, the one regressing case, +1.8%), 200 (`Medium`/`Large`, -17.9%), 300 (`XLarge`/`2XLarge`, -28.5%). Left `_MASK_CHUNK` untouched — a smaller chunk size would likely flip the `num_select=100` case too, but that trades off against loop/kernel-launch overhead, a separate thing from what this PR is about.

**Validation**

- `tests/models/test_postprocess.py`: 29/29 passed, CPU and CUDA both pinned.
- `pre-commit run --all-files`: 19 hooks clean, `mypy --strict` included.
- CPU, K=300 at 1920x1080 (largest realistic case): latency down ~12%, fresh-process peak RSS down ~21% (~360 MB).
- CUDA, same case: latency down ~17%, peak allocated memory down ~29%.
- Real model + real image (`RFDETRSegNano`, 2048x1024, K=100 masks): CPU latency down ~12%, CUDA latency roughly flat (~1% faster), CUDA peak memory up ~3.75% (the residual above), identical output hash on both devices, every run.
- Small outputs (~19 MB) come out as a small, consistent win on both devices now — an earlier version of this without the `del` fix measured that size as a wash.
- Patch applies clean on current `develop`.

Small one, but it's the kind of thing that adds up once you're returning masks for a full batch. Hope this helps!


Borda:
# Segmentation mask postprocess benchmark results

## Run

- Date: 2026-08-20
- Harness: [`benchmarks/postprocess_mask_memory.py`](../../benchmarks/postprocess_mask_memory.py)
- Device: CPU (`macOS-26.6.1-arm64`, Python 3.12.14, PyTorch 2.12.1)
- Seed: `0`
- Mask head: `192×192`; output: `128×128`
- Chunk size: `32`
- Warmup/repeats: `1/3`
- Synchronization: enabled for the measured device
- Memory metric: process high-water RSS via `resource.getrusage`

Command, run once per variant and K:

```bash
uv run --no-sync python benchmarks/postprocess_mask_memory.py \
    --variant {legacy|preallocated} --k {100|200|300} \
    --height 128 --width 128 --warmup 1 --repeats 3
```

## Results

|   K | Variant      | Median (ms) | Min (ms) | Max (ms) | Peak RSS (bytes) | Output true-count |
| --: | ------------ | ----------: | -------: | -------: | ---------------: | ----------------: |
| 100 | legacy       |    1.494167 | 1.311333 | 2.255917 |      238,141,440 |           818,659 |
| 100 | preallocated |    1.131625 | 1.073792 | 1.280542 |      237,699,072 |           818,659 |
| 200 | legacy       |    3.850459 | 2.791000 | 4.481791 |      274,710,528 |         1,638,880 |
| 200 | preallocated |    2.871500 | 2.115625 | 5.078166 |      270,368,768 |         1,638,880 |
| 300 | legacy       |    4.937416 | 4.118791 | 9.184458 |      305,790,976 |         2,457,720 |
| 300 | preallocated |    5.928208 | 3.215000 | 6.731500 |      300,122,112 |         2,457,720 |

## Findings

- All six invocations exited successfully.
- Output checksums matched between legacy and preallocated variants for every K value.
- In this bounded CPU run, preallocation used lower observed peak RSS for all three K values.
- Timing is noisy at this small output size: preallocation was faster at K=100 and K=200, but slower at K=300. Treat these values as a smoke comparison, not an end-to-end performance claim.

## Limits and follow-up

- This run does not measure CUDA allocator memory or CUDA timing; CUDA was unavailable on the host.
- It does not reproduce the PR-scale `1080×1920` case. The corresponding command is:

```bash
uv run --no-sync python benchmarks/postprocess_mask_memory.py \
    --variant {legacy|preallocated} --k {100|200|300} \
    --height 1080 --width 1920 --warmup 1 --repeats 3 --device cuda
```

- Run the PR-scale CUDA matrix on a CUDA host before using the results as maintainer acceptance evidence.

Borda (approved):

PR #1378, opened 2026-08-21
feat(config): add `eval_batch_size` for eval loaders

Borda:
* Add `TrainConfig.eval_batch_size` (`int | None`, default `None`) with a validator rejecting values below 1; `None` inherits the resolved training batch size, so no existing run changes shape.
* Add `RFDETRDataModule._resolve_eval_batch_size()` and route the validation, test and predict dataloaders through it; the training dataloader and its `grad_accum_steps` alignment padding are untouched.
* An explicit `eval_batch_size` is returned without consulting `batch_size`, so it stays usable on the `batch_size="auto"` path where `_resolve_batch_size()` raises; leaving it unset preserves that error for eval loaders too.
* Change the `grad_accum_steps` default from 4 to 1, moving the default effective batch size from 16 to 4. This is a training-semantics change, not a throughput-only one: the optimization trajectory differs and convergence may differ with it. Set `grad_accum_steps=4` explicitly to restore the previous behaviour. Runs using `batch_size="auto"` are unaffected because the auto-batch probe overwrites the field with its own recommendation.
* Rewrite the batch and gradient-accumulation guidance across `docs/learn/train/training-parameters.md`, `docs/faq.md`, `docs/learn/train/index.md` and `docs/root-static/llms-full.txt` to recommend raising `batch_size` first and accumulating only when memory caps it, and to correct the published default that still read 4.
* Add config tests for the new validator and datamodule tests covering default inheritance, an explicit override applied to all three eval loaders, and the training loader's independence from it.


LeMinhNgan:
@Borda , very interesting that you set default ``grad_accum_steps= 1`` and increase ``batch_size`` 4 times to optimize memory usage. Curious about VRAM usage during validation when keeping ``eval_batch_size=batch_size`` when we set it None. The current pipeline makes train and validation using same VRAM amount. 
You note "so it can often run larger batches than training" means we can set `eval_batch_size=2*batch_size `for example?. I hope I can shorten the non-value-added validation time with the change. It'll be good if you can modify ``gpu_memory_progress_bar.py`` to show the `free_mem` during validation as we are doing for epoch-training, then we can maximize the ``eval_batch_size`` easily

Borda:
> very interesting that you set default `grad_accum_steps= 1` and increase `batch_size` 4 times to optimize memory usage

All of it is optimization-driven; from my profiling run, it came out that accumulation just makes it slower, so if there is no real need to make the effective batch larger, it is better not to use it...

> I hope I can shorten the non-value-added validation time with the change.

The capabilities and flexibility shall remain; we just pay a bit with the default setup so the user gets faster training out of the box 🦝

PR #1384, opened 2026-08-21
docs(cookbooks): add COCO2017 full-dataset training cookbook

Borda:
Demonstrates training RFDETRNano for 40 epochs on the full public COCO2017 dataset on a single Colab GPU, using shipped TrainConfig defaults that already carry the recent training-throughput optimizations.


docs/cookbooks/cards.yaml
```diff
@@ -1,4 +1,10 @@
 cards:
+  - href: train-coco2017/
+    name: "Train RF-DETR Nano on COCO2017"
+    labels: [TRAINING, PYTORCH LIGHTNING]
```
Borda:
```suggestion
    labels: [TRAINING, COCO2017, PYTORCH LIGHTNING]
```

PR #1395, opened 2026-08-22
fix(augment): align Kornia GaussianBlur/GaussNoise defaults with albumentations

### LINKED ISSUE #1351: Kornia GaussianBlur and GaussNoise defaults diverge from the Albumentations defaults they stand in for
### Search before asking

### Bug

Two Kornia factory defaults do not match the Albumentations defaults they stand in for, so switching `augmentation_backend` silently changes augmentation strength for the same `aug_config`. One of them is reachable through a shipped preset.

**1. `GaussianBlur` sigma.** `AUG_INDUSTRIAL` contains `{"GaussianBlur": {"blur_limit": 3, "p": 0.3}}` with no `sigma`, so both backends fall back to their own default:

```
GPU (kornia)         sigma = (0.1, 2.0)
CPU (albumentations) sigma = (0.5, 3.0)
```

The GPU path blurs noticeably less. The comment at the call site claims the opposite:

```python
# Match the CPU albumentations default sigma range while allowing an explicit override via config.
sigma_range = params.get("sigma", (0.1, 2.0))
```

`(0.1, 2.0)` is Kornia's convention, not Albumentations'. The doc table in `aug_configs.py` repeats it as `sigma=(0.1, 2.0)`.

**2. `GaussNoise` std.** With `{"GaussNoise": {"p": 0.3}}` and no `std_range`:

```
GPU (kornia)         std       = 0.05      (fixed)
CPU (albumentations) std_range = (0.2, 0.44)
```

Roughly a 4-9x difference in noise magnitude. The warning emitted here also reports rf-detr's own default as if it were Albumentations':

```
GPU augmentation (Kornia) uses fixed std=0.050 for GaussianNoise ...
CPU augmentation (albumentations) samples from [0.010, 0.050].
```

Albumentations would sample from `(0.2, 0.44)` for that config, not `[0.01, 0.05]`, so the line that exists to make the divergence visible understates it.

### Expected behavior

For the same `aug_config`, an unspecified parameter should resolve to the same range on both backends, and the divergence warnings should describe what Albumentations would actually do.

### Minimal reproducible example

```python
import albumentations as A
from rfdetr.datasets.kornia_transforms import _make_gaussian_blur

# The GaussianBlur entry from AUG_INDUSTRIAL
print(_make_gaussian_blur({"blur_limit": 3, "p": 0.3}).flags["sigma"])  # (0.1, 2.0)
print(A.GaussianBlur(blur_limit=3, p=0.3).sigma_limit)                  # (0.5, 3.0)
```

### Environment

- RF-DETR: `develop` at `b18c9ca`
- albumentations 2.0.8, kornia 0.8.3, torch 2.6.0
- CPU only; no GPU or weights needed

### Additional

A caveat I would rather state than gloss over: `pyproject.toml` allows `albumentations>=1.4.24,<3.0.0`, and I verified these numbers against 2.0.8 only. If 1.4.x used different defaults then the correct fix is version-aware rather than a straight constant swap, and I have not checked that.

There is also a judgment call here that is yours rather than mine: aligning the defaults changes augmentation strength for anyone currently training on the Kornia backend with these presets. The alternatives are to match Albumentations (consistent, but changes behaviour), or to keep the values and correct the comment, the doc table and the warning so they describe what actually happens. I lean towards matching, since the GPU backend is documented as a drop-in for the CPU one, but I would rather ask than pick for you.

Related: #1349 was the same class of problem in `CLAHE`.

adhavan18:
## What

Aligns two Kornia augmentation-factory defaults with the Albumentations defaults they stand in for. Closes #1351.

## Why

For the same `aug_config`, switching `augmentation_backend` silently changed augmentation strength, because two Kornia defaults did not match Albumentations':

**GaussianBlur sigma.** `AUG_INDUSTRIAL` sets `{"GaussianBlur": {"blur_limit": 3, "p": 0.3}}` with no `sigma`, so both backends fall back to their own default:

```
GPU (kornia)         sigma = (0.1, 2.0)   <- Kornia's convention
CPU (albumentations) sigma = (0.5, 3.0)   <- what the config actually means
```

The call-site comment claimed it "Match[es] the CPU albumentations default", which was the opposite of what it did.

**GaussNoise std_range.** Defaulted to rf-detr's own `(0.01, 0.05)` against the Albumentations default `(0.2, 0.44)` on the same 0-1 image scale, 4-9x weaker. The divergence warning also reported rf-detr's own `[0.01, 0.05]` as if it were Albumentations', hiding the gap.

Verified against the pinned `albumentations==2.0.8`:

```python
>>> A.GaussianBlur().sigma_limit
(0.5, 3.0)
>>> A.GaussNoise().std_range
(0.2, 0.44)
```

## Changes

- `_make_gaussian_blur` default sigma `(0.1, 2.0)` -> `(0.5, 3.0)`, comment corrected.
- `_make_gauss_noise` default std_range `(0.01, 0.05)` -> `(0.2, 0.44)`. An explicit range still wins, and the warning text is unchanged since it is still correct for a non-degenerate explicit range.
- Doc table in `aug_configs.py` updated.

This changes only the *default* used when a config omits the value; any config that sets `sigma`/`std_range` explicitly is unaffected. It follows the same "align to Albumentations" direction accepted for the sibling scalar-`clip_limit` fix (#1350).

## Testing

Three regression tests pin both defaults to the live Albumentations values. The two that assert the defaults fail on `develop` and pass here:

```
develop:     unspecified GaussianBlur sigma is (0.1, 2.0) ... FAILED
             unspecified GaussNoise std is 0.05 ...          FAILED
this branch: 8 passed (Gauss* subset)
```

Full augmentation suite: `55 passed`. `ruff check` and `ruff format --check` clean.

### DIFF
--- a/src/rfdetr/datasets/aug_configs.py
+++ b/src/rfdetr/datasets/aug_configs.py
@@ -84,7 +84,7 @@ or torchvision defaults. Install it with ``pip install 'rfdetr[augment]'``.
 | ``ColorJitter`` | ``K.ColorJiggle`` | Same multiplicative semantics |
 | ``ToGray`` | ``K.RandomGrayscale`` | Grayscale, 3 channels; only ``p`` honored, method/num_output_channels ignored |
 | ``RandomBrightnessContrast`` | ``K.ColorJiggle`` | ``brightness_limit`` / ``contrast_limit`` direct |
-| ``GaussianBlur`` | ``K.RandomGaussianBlur`` | ``blur_limit`` rounded up to odd; ``sigma`` default to alb. (0.5, 3.0) |
+| ``GaussianBlur`` | ``K.RandomGaussianBlur`` | ``blur_limit`` rounded up to odd; ``sigma`` mirrors Albumentations |
 | ``GaussNoise`` | ``K.RandomGaussianNoise`` | Upper bound of ``std_range`` used as fixed std |
 | ``Blur`` | ``K.RandomBoxBlur`` | Box blur; ``blur_limit`` rounded up to odd, pair collapses to its upper bound |
 | ``Sharpen`` | ``K.RandomSharpness`` | ``sharpness = 1.0 + alpha`` (1.0-pivoted); ``lightness``/``method`` ignored |
--- a/src/rfdetr/datasets/aug_configs.py
+++ b/src/rfdetr/datasets/aug_configs.py
@@ -84,7 +84,7 @@ or torchvision defaults. Install it with ``pip install 'rfdetr[augment]'``.
 | ``ColorJitter`` | ``K.ColorJiggle`` | Same multiplicative semantics |
 | ``ToGray`` | ``K.RandomGrayscale`` | Grayscale, 3 channels; only ``p`` honored, method/num_output_channels ignored |
 | ``RandomBrightnessContrast`` | ``K.ColorJiggle`` | ``brightness_limit`` / ``contrast_limit`` direct |
-| ``GaussianBlur`` | ``K.RandomGaussianBlur`` | ``blur_limit`` rounded up to odd; ``sigma=(0.1, 2.0)`` |
+| ``GaussianBlur`` | ``K.RandomGaussianBlur`` | ``blur_limit`` rounded up to odd; ``sigma`` defaults to the albumentations (0.5, 3.0) |
 | ``GaussNoise`` | ``K.RandomGaussianNoise`` | Upper bound of ``std_range`` used as fixed std |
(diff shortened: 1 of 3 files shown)


Borda (dismissed):

adhavan18:
Thanks for picking this up and polishing it @Borda, the reworded GaussNoise note is exactly right: bounds aligned, but Kornia's fixed std is not the CPU path's sampled range, so "same default, not same distribution" is the precise claim.

Happy to take anything else you'd like on it, otherwise I'll leave it in your hands.

Borda (approved):

PR #1406, opened 2026-08-26
Proposal: avoid unsupported TorchScript on Python 3.14

cdeil:
Hi! While checking RF-DETR on Python 3.14, I saw this warning during a normal top-level import with PyTorch 2.13:

```text
DeprecationWarning: `torch.jit.script` is not supported in Python 3.14+ and may break. Please switch to `torch.compile` or `torch.export`.
```

Would you be interested in handling it with the conditional eager fallback proposed here?

## Evidence

Importing `rfdetr` executes four module-level `torch.jit.script` calls:

- [`batch_dice_loss`](https://github.com/roboflow/rf-detr/blob/6674d858/src/rfdetr/utilities/box_ops.py#L210) and [`batch_sigmoid_ce_loss`](https://github.com/roboflow/rf-detr/blob/6674d858/src/rfdetr/utilities/box_ops.py#L234)
- [`dice_loss`](https://github.com/roboflow/rf-detr/blob/6674d858/src/rfdetr/models/criterion.py#L315) and [`sigmoid_ce_loss`](https://github.com/roboflow/rf-detr/blob/6674d858/src/rfdetr/models/criterion.py#L339)

With the warning promoted to an error, current `develop` stops at the first site:

```console
$ python -c "import warnings; warnings.filterwarnings('error', message=r'.*torch\\.jit\\.script.*', category=DeprecationWarning); import rfdetr"
...
  File "src/rfdetr/utilities/box_ops.py", line 210, in <module>
    batch_dice_loss_jit = torch.jit.script(batch_dice_loss)
DeprecationWarning: `torch.jit.script` is not supported in Python 3.14+ and may break.
```

The existing [dependency-resolution matrix includes Python 3.14](https://github.com/roboflow/rf-detr/blob/6674d858/.github/workflows/ci-deps-resolution.yml#L83-L97), but it only performs an install-plan dry run. The executable [CPU test matrix currently stops at Python 3.13](https://github.com/roboflow/rf-detr/blob/6674d858/.github/workflows/ci-tests-cpu.yml#L28-L41), so it cannot observe an import-time warning. A fresh install using the CPU workflow's command currently resolves PyTorch 2.13 and reproduces it.

## Proposed behavior

- Keep the existing scripted aliases on Python 3.10–3.13.
- Use the original eager loss functions on Python 3.14+, where TorchScript is unsupported.
- Add Ubuntu/Python 3.14 to the CPU matrix and a targeted subprocess import regression test.
- Advertise Python 3.14 in the package classifier and contributor docs now that the real suite executes there.

This does not appear to be a performance issue at import time: seven fresh-process imports were effectively unchanged (1.982 s vs. 1.970 s median), and the four top-level scripting calls accounted for roughly 8 ms. A small CPU loss microbenchmark also showed identical outputs and only microseconds of eager overhead (Dice: 21.9 vs. 20.0 µs; sigmoid CE: 44.8 vs. 44.7 µs). The motivation is avoiding an explicitly unsupported path before it becomes a failure.

## Validation

- Red: the targeted warning-as-error import fails on current `develop` at `box_ops.py:210`.
- Green: 4,394 CPU tests passed and 80 skipped on CPython 3.14.6 with PyTorch 2.13.0.
- Green: `pre-commit run --all-files` passed in the documented Python 3.10 tooling environment.

Would this conditional eager fallback be the compatibility policy you prefer, or would you rather remove or replace these TorchScript aliases more broadly?

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -115,7 +115,7 @@ rf-detr/
 
 ## Development Environment Setup
 
-RF-DETR uses **`uv`** as the package manager for dependency management. Ensure you have Python >=3.10 installed (supports 3.10, 3.11, 3.12, 3.13).
+RF-DETR uses **`uv`** as the package manager for dependency management. Ensure you have Python >=3.10 installed (supports 3.10, 3.11, 3.12, 3.13, 3.14).
 
 ### Installing uv
 
@@ -282,7 +282,7 @@ Tests marked with `@pytest.mark.gpu` are excluded from CPU CI workflows and run 
 Our continuous integration tests run on:
 
 - **Operating Systems:** Ubuntu, Windows, macOS
-- **Python Versions:** 3.10, 3.11, 3.12, 3.13
+- **Python Versions:** 3.10, 3.11, 3.12, 3.13, 3.14
 - **CPU Workflow:** `pytest -m "not gpu"` - Runs on all OS/Python combinations
 - **GPU Workflow:** `pytest -m gpu` - Runs separately on GPU infrastructure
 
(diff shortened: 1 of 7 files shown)


cdeil:

Borda (dismissed):

Borda (dismissed):

Borda (approved):

Issue #1058, opened 2026-05-22
TPU Training Support


Borda:
## Summary

RF-DETR can be made trainable on Google TPUs (v3, v5e, v6e) with a small set of targeted fixes.
PyTorch Lightning's `XLAStrategy` handles all TPU infrastructure automatically — distributed init,
data loading, checkpointing. The remaining work is making the RF-DETR model code XLA-compatible
and fixing two known performance bottlenecks.

All changes are device-gated — no GPU or CPU behaviour changes.

---

## Motivation

- Free TPU quota on Colab (v5e-1, v6e-1) and Kaggle (v3-8) makes TPU training accessible to the
    community without GPU costs.
- TPU v5e/v6e offer significantly higher bf16 throughput than comparable GPU instances.
- Several community members have requested TPU support (see #XXX).

---

### Phase 0 — Test infrastructure (prerequisite)

**Add `@pytest.mark.tpu` pytest marker and update CI filters.**

- Register marker in `pyproject.toml` alongside existing `gpu` and `flaky` markers.
- Update CPU CI filter from `-m "not gpu"` to `-m "not gpu and not tpu"` so TPU tests are skipped
    in standard CI but remain runnable explicitly with `-m tpu`.

---

### Phase 1 — XLA compatibility (7 tasks)

Tasks 1.1, 1.2, 1.3, 1.6, 1.7 are independent and can be developed in parallel.

#### Task 1.1 — XLA branch in `_bilinear_grid_sample` (`tensors.py:228`)

`F.grid_sample` backward is a known CPU-fallback op on XLA. The gather-based path already used
for MPS is fully XLA-compatible. One-line fix:

```python
# before
if input.device.type != "mps":
# after
if input.device.type not in ("mps", "xla"):
```

#### Task 1.2 — Remove `.item()` from `num_boxes` (`criterion.py:497`)

`.item()` forces a device-to-host sync, cutting the XLA lazy graph on every training step.
Scalar tensor division works identically on CPU, CUDA, and XLA.

```python
# before
num_boxes = torch.clamp(num_boxes / get_world_size(), min=1).item()
# after
num_boxes = torch.clamp(num_boxes / get_world_size(), min=1)  # keep as tensor
```

**Note:** `dice_loss` and `sigmoid_ce_loss` are `@torch.jit.script` functions with
`num_masks: float` signatures — these must be updated to `num_masks: torch.Tensor` in the same PR.

[...]

PR #1432, opened 2026-09-04
fix(models): route nearest `point_sample` through the backend-agnostic gather path

JESUSROYETH:
## Summary

`point_sample(mode="nearest")` delegates to `F.grid_sample`, which on XLA lowers to an `aten::grid_sampler_2d` host fallback. `SetCriterion.loss_masks` calls it once per matched output layer to sample ground-truth mask labels, and `HungarianMatcher.forward` calls it at that same cadence for the segmentation cost matrix. So, every segmentation training step leaves the device several times over.

This PR routes the nearest branch through `_nearest_grid_sample`, the same counterpart the bilinear branch already has in `_bilinear_grid_sample`. CUDA and CPU don't change — off MPS/XLA the helper still delegates to `F.grid_sample`'s fused kernel, which is faster there. Only MPS/XLA take the gather path.

Other modes and unsupported padding modes still reach `F.grid_sample` unchanged.

## Correctness

Ties are rounded with `torch.round` (round-half-to-even), which matches `F.grid_sample` for random grids and for most constructed exact ties. `floor(x + 0.5)` is a worse substitute, but it's not the opposite in every case — it disagrees with round-half-to-even at half of the exact ties, not all of them (e.g. it agrees with `torch.round` at -2.5, -0.5 and 1.5, and disagrees at -1.5, 0.5 and 2.5). Random grids never land on a tie, so the parity tests use grids built to fall exactly on `.5` in source-index space:

| rounding | agrees with `F.grid_sample` at exact ties |
|---|---|
| `torch.round` (half-to-even) | 4 of 4 padding/align combinations |
| `floor(x + 0.5)` | 0 of 4 |

Against `F.grid_sample(mode="nearest")` the gather path is bit-identical (`torch.equal`, not a tolerance) across 480 randomized cases, spanning both padding modes, both `align_corners` settings, and float32/float64, plus the constructed tie grids above. Parity also covers batch size > 1 with non-square `H`/`W` and grid dimensions — every real caller passes a batch dimension equal to the matched-instance count, routinely > 1, not the `1` used in the other cases. It also covers float16/bfloat16: `HungarianMatcher` casts target masks to `pred_masks_logits`'s dtype before this call, so autocast training reaches this path in low precision too, not just float32.

This does not hold for every possible input, and we need to be honest about that. `F.grid_sample`'s own compiled kernel breaks its own round-half-to-even convention at specific (size, position) combinations — e.g. it rounds the exact tie `x=0.5` up to index 1 at width=673, and down to index 0 at width=96. This same divergence is already documented and corrected for on CPU/CUDA, in `SetCriterion._sample_target_masks_at_points` (`src/rfdetr/models/criterion.py`), which falls back to the real kernel for tied points because `torch.round` can't be relied on to match it everywhere. On XLA/MPS that fallback now goes through this same gather path, so it no longer reaches the kernel's original tie value for that rare case. The divergence is reproduced for the width=673 case in `tests/utilities/test_tensors.py::TestNearestGridSampleParity::test_known_kernel_tie_divergence_at_width_673`.

Like its bilinear counterpart, the gather path also doesn't validate a malformed `grid` (mismatched batch size, or a non-2 trailing dimension) the way `F.grid_sample` does. Both helpers assume the well-formed shapes `point_sample`'s real callers always pass.

## Measurements

Cloud TPU v6e-1, `RFDETRSegNano` segmentation training, 5 optimizer steps at 252 px, batch 2, one fresh process per arm, two alternating pairs:

| | base | patched |
|---|---:|---:|
| `aten::grid_sampler_2d` | 50, 50 | **0, 0** |
| uncached compilations | 26, 26 | **16, 16** |
| scalar host reads | 110, 110 | 110, 110 |
| fit time | 352.24 s, 356.94 s | **287.36 s, 287.73 s** |

Fallback count and compile count are deterministic and identical in both pairs, fit time improved 18.42% and 19.39% with disjoint envelopes.

That time figure is not a training-speed claim. A 5-step fit is dominated by compilation (16–26 compiles for 5 steps), so most of the improvement is just the 10 avoided compilations, and those amortize away on a long run. What stays is the removed host fallback and the graph cuts around it. Scalar host reads are unchanged at 110, which confirms this change does one thing.

Refs #1058, whose Phase 2a criterion "zero `aten::` fallback ops" currently fails for segmentation.


src/rfdetr/models/heads/segmentation.py
```diff
@@ -386,8 +386,19 @@ def point_sample(input: Tensor, point_coords: Tensor, **kwargs: Any) -> Tensor:
...
+    elif mode == "nearest" and not kwargs and padding_mode in ("zeros", "border"):
+        # Same reasoning as the bilinear branch: F.grid_sample lowers to an aten::grid_sampler_2d
+        # host fallback on XLA, so the gather path keeps mask-label sampling on device. CUDA/CPU
+        # still get the fused kernel from inside the helper. Exact ties retain the helper's documented
+        # kernel-dependent backend difference.
+        output = _nearest_grid_sample(
+            input,
+            grid,
+            padding_mode=padding_mode,
+            align_corners=align_corners,
```
Borda:
```suggestion
            align_corners=align_corners
```

Borda (approved):

PR #1434, opened 2026-09-04
docs(export): link a community RF-DETR Nano Android (LiteRT GPU) example

john-rocky:
Thanks for the TFLite export section and its honest experimental note. This adds a short community-example subsection at the end of it: RF-DETR Nano converted to LiteRT (.tflite) and running fully on an Android GPU through the LiteRT CompiledModel API, with a Kotlin camera app.

- The conversion did not go through the `onnx2tf` route above; it used `litert-torch` plus a two-graph split (query selection on the host) so both graphs stay on the GPU. The subsection says so, to keep it apart from `model.export(format="tflite")`.
- Pixel 8a: live camera at about 9 fps (~110 ms per frame); detections match PyTorch at IoU 0.98–0.99.
- Apache-2.0 weights only (Nano); the zoo section links back to this repo.
- The paragraph is on one line for the mdformat no-wrap check.

Zoo section: https://github.com/john-rocky/LiteRT-Models#rf-detr-nano — weights: https://huggingface.co/litert-community/RF-DETR-Nano-LiteRT

Happy to move it (README, FAQ) or drop it if it does not fit the docs. Thanks again.

### DIFF
--- a/docs/learn/export.md
+++ b/docs/learn/export.md
@@ -404,6 +404,10 @@ boxes = interpreter.get_tensor(boxes_detail["index"])
 labels = interpreter.get_tensor(labels_detail["index"])
 ```
 
+### Community example: RF-DETR Nano on an Android GPU (LiteRT)
+
+[LiteRT-Models › RF-DETR Nano](https://github.com/john-rocky/LiteRT-Models#rf-detr-nano) is a community conversion of RF-DETR Nano to LiteRT (`.tflite`) with a Kotlin camera app. It was converted with `litert-torch` rather than the `onnx2tf` route above, and split into two graphs (the two-stage query selection runs on the host between them) so both graphs run fully on the phone GPU through the LiteRT CompiledModel API; on a Pixel 8a it runs live camera at ~9 fps (~110 ms/frame) with detections at IoU 0.98–0.99 against PyTorch.
+
 ## OpenVINO IR Export
 
 OpenVINO IR (Intermediate Representation) is a proprietary model format used by the OpenVINO Toolkit to optimize and deploy deep learning models.


Borda:
@john-rocky looks nice! How about adding this to export directly from RFDETR?

john-rocky:
Thanks, @Borda. Glad you like it, and yes, I'd be happy to bring it into `export()` itself. I'd do that as a separate PR against `develop`, so this one-line docs pointer stays independent (the branch is rebased on `develop` after the OpenVINO section landed, so it can go in or be dropped on its own).

The shape mirrors the OpenVINO route: `format="litert"` behind a new `rfdetr[litert]` extra pinned to `litert-torch`, an `export/_litert/` package with the same `ImportError` guard naming the extra, the dispatch next to the OpenVINO branch, tests under `tests/export/`, and a docs section. The conversion goes `torch.export` → `.tflite` directly, with no ONNX or TensorFlow step in between, and produces the same `(boxes, logits)` outputs as the current TFLite route.

One thing to say up front: the graph in the linked app is not a plain convert, and a plain convert does not go through today. I re-checked on `develop` with litert-torch 0.9.4, and it stops in the deformable-attention lowering (the per-level `split` in `ms_deform_attn_core_pytorch`). The app's graph gets through with a few export-time rewrites (deformable-attention sampling as a gather-free bilinear matmul, the windowed DINOv2 attention kept at ≤4-D, fp16-safe LayerNorms for the phone GPU), and with them the single-graph `.tflite` matches PyTorch at corr 1.0 on CPU. The GPU additionally needs a two-graph split with the top-300 query selection on the host, since the `TopK`/`Gather` left in the single graph have no GPU kernels. So I'd go in two steps: the single-graph `format="litert"` first (rewrites included; CPU / XNNPACK, full parity), then the GPU split plus a small `LiteRTInference` helper for the host step as an opt-in follow-up.

If you'd rather see the plan as an issue first, as the contributing guide asks for new features, I'll open that instead. Thanks again for the invitation.

john-rocky:
Thanks again for the nudge, @Borda — the `export()` route is up as #1459. Two things came out simpler than I wrote above: a plain convert of `develop` needed only one rf-detr change (the single-level `split` in `ms_deform_attn_core_pytorch`, which litert-torch 0.9.4 cannot lower), and with that alone the single-graph `.tflite` matches eager PyTorch on CPU (boxes ~1e-7, logits ~2e-5 on the top detections), so the GPU-oriented rewrites from the app stayed out of the PR. The phone-GPU split is the follow-up if it is wanted. This docs PR stays independent; happy to drop it if the new section makes it redundant.

Borda:
closing in favor of #1459

PR #1435, opened 2026-09-05
refactor(datasets): outsource synthetic generator

Borda:
- Delete `src/rfdetr/datasets/synthetic.py`; its only importers were RF-DETR's own test fixtures, and `fuse_augmentations.data` generates the same shape datasets in COCO or YOLO layout for detection, segmentation and OBB
- `tests/conftest.py` builds both dataset fixtures through `generate_dataset`, with the previously duplicated `val`-to-`valid` rename and `test`-split clone folded into a shared `build_synthetic_dataset` helper
- Add `sparsify_category_ids`, which re-encodes the dense category ids the package writes into `1, 3, 5, …` so the fixtures keep exercising rf-detr's gap-tolerant `cat2label` remapping
- Pass `SYNTHETIC_DATASET_SEED` to every generation call: `generate_dataset` draws from fresh entropy when `seed` is unset, which would make the session-scoped fixtures differ between runs
- Delete `tests/datasets/test_synthetic.py`; the generator's own behaviour is now the package's to test
- Pin `fuse-augmentations==0.10.0.dev0` in the `tests` dependency group, exactly rather than as a range because the project only publishes dev releases so far
- Document the removal as a breaking change in `CHANGELOG.md` and add an `Upgrade 1.10 → 1.11` section to the migration guide covering the before/after call, the dense-id difference and the added `rectangle` shape


Borda:
Bottom line: we added synth data only for testing, so keeping it would just be dead weight. Also, with this package, we can train on synthetic keypoints...

PR #1451, opened 2026-09-10
test(training): cover `pack_targets` round-trip for segmentation masks

JESUSROYETH:
## What this closes

My own #1399 ("pack_targets") shipped with this in its "Not covered" section:

> Segmentation and keypoints. The 114 → 9 count is for detection targets. Those heads add a field, making it 130 → 10, and the packer handles any same-keyed tensor field, but neither was measured — no throughput number and no parity run for them.

This closes the parity half for segmentation: a correctness test for the `masks` field, verified against both synthetic and real data. Keypoints, and a throughput measurement for segmentation, are deliberately separate follow-ups — the throughput question turned out to need more care than fits in this PR, and I'd rather keep that methodology debate off a test that has none.

## No source change

`pack_targets()` ([`tensors.py#L666`](https://github.com/roboflow/rf-detr/blob/75c25038/src/rfdetr/utilities/tensors.py#L666)) is already generic over any same-keyed tensor field — it iterates the batch's key set with no field-name branching anywhere in the pack or rebuild path. `masks` was never a special case; it just hadn't been driven through a real segmentation batch by a test.

The other gap #1399 flagged for masks specifically — a transient doubling of the field's memory while `transfer_batch_to_device` cloned it after moving the whole packed object — was already fixed independently, by merged #1405. `PackedTargets.to_list()` now materializes each packed field directly into its destination-device tensor per sample, and its own docstring names exactly this case ("negligible for boxes and labels but can be large for segmentation masks").

## What's added

- `TestTrainDataloader.test_pack_targets_round_trips_segmentation_masks_bit_identically`: builds three synthetic segmentation samples (`RFDETRBaseConfig(segmentation_head=True)`, `masks` as `torch.bool` of shape `(num_instances, H, W)`, matching what `rfdetr.datasets.coco.py` actually produces) with 1, 0, and 3 instances — the middle, zero-instance sample mirrors `TestPackedTargets.test_a_sample_with_no_instances_survives_the_round_trip` in `tests/utilities/test_tensors.py`, which already pins that same "must not collapse into a neighbour's rows" invariant for `boxes`/`labels` but never through a real segmentation batch's `masks` field. Drives all three through the real `RFDETRDataModule.train_dataloader().collate_fn`, and asserts the batch still packs and that `masks`/`boxes`/`labels`/`area`/`iscrowd` round-trip bit-identically via `to_list()` — the exact method `transfer_batch_to_device` calls on the real training path, not a related-but-different iteration method.
- A doctested helper building that sample shape, reused by the test.
- A one-line CHANGELOG entry.

## Evidence

The test is not vacuous: I mutated `pack_targets()` to bit-flip the `masks` field specifically after packing, confirmed the test fails with exactly the expected message, then reverted (confirmed via a clean `git diff` on the source file — it ships unchanged). That mutation is what proves `masks` specifically discriminates. Separately, to confirm the zero-instance sample is actually exercised rather than silently skipped, I injected an off-by-one into the packed-offset arithmetic that only miscounts zero-element shapes — it fails on `area` (alphabetically first among the packed keys) before reaching `masks`, which proves the zero-instance route runs real offset/shape bookkeeping rather than being bypassed, not that the `masks` assertion itself is what catches it. Reverted the same way.

The committed test's data is deliberately synthetic (self-contained, no network, no external fixture) — driven through the real collate path, but the tensors themselves are fabricated. Separately confirmed the same property against real data: two real COCO photos with their real, human-drawn instance-segmentation polygons, read through the real `YoloDetection` reader — real JPEG decode, real polygon-to-mask rasterization — materialized with `packed.to_list()` (the same production method the committed test uses), and packed through the real collate path. Both round-trip bit-identically; instance count per image varies run to run (unseeded augmentation clips a different subset near crop boundaries each time — observed 1 and 6, and separately 1 and 8, across repeats), which doesn't affect the bit-identity result either way.

## Testing

Full CPU suite, CI's exact marker set (`-m "not gpu and not coco17 and not e2e_coreml and not e2e_executorch and not e2e_openvino and not e2e_roboflow and not xla and not tpu"`, matching `ci-tests-cpu.yml`): every failure on the patched tree reproduces identically on unpatched `develop`. None touches `pack_targets`, `PackedTargets`, segmentation, or the dataloader/collate path.

`pre-commit run --all-files`, including `mypy --strict`: 19/19 hooks pass.

## Known collision

My own #1450 (pad training targets to a fixed row count for XLA, still open) adds tests near the same class in `tests/training/test_module_data.py`. Checked with `git apply --check` against #1450's current head: `CHANGELOG.md` applies cleanly, the test file doesn't. Whichever of the two merges second will need a short rebase; it's a few lines of context, not a logical conflict.


Borda:
@JESUSROYETH, pls do not force push to PR once it's not just a draft; it resets my review process and discards all my changes...

Borda (approved):

PR #1231, opened 2026-07-23
feat(export): fp16 engine arg + export cookbooks

Borda:
This pull request adds new documentation and example notebooks for exporting RF-DETR models to TensorRT and ExecuTorch formats, and updates related documentation to clarify export options and improve discoverability. It also adjusts the pre-commit license header insertion to avoid running on documentation files.

**New export cookbooks and documentation:**

* Added two new example notebooks: `export-tensorrt.ipynb` (demonstrates exporting to TensorRT and running inference on NVIDIA GPUs) and `export-executorch.ipynb` (demonstrates exporting to ExecuTorch and running inference on mobile/edge devices). Both include full, step-by-step guides with visualization. [[1]](diffhunk://#diff-4fc5369c2452d3da9c1c38d27a7a547410cd89087330be2aac4729b4d6992191R1-R289) [[2]](diffhunk://#diff-d68a1babaec3cd0e0f696d23d0154e5329f1d91c61e41fb7b819e3f47a52dad6R1-R277)
* Registered the new cookbooks in `cards.yaml` for discoverability, including descriptions, labels, and versioning metadata.
* Updated the cookbooks table in `NOTES.md` to include the new export notebooks and their versions.

**Export documentation improvements:**

* Clarified the `export()` method documentation to document the new `fp16` parameter for TensorRT export and improved the formatting of the parameters table. [[1]](diffhunk://#diff-76bf72360ae4b094d0fe12835eae94e7fa98534d371672cb78af21088eae67e8L64-R64) [[2]](diffhunk://#diff-76bf72360ae4b094d0fe12835eae94e7fa98534d371672cb78af21088eae67e8R80)

**Tooling and configuration:**

* Updated `.pre-commit-config.yaml` so that the license header insertion hook excludes files in the `docs/` directory instead of `notebooks/`, preventing license headers from being inserted into documentation files.

PR #1054, opened 2026-05-21
chore(tflite): rewrite GridSample to Gather for TFLite correctness

Borda:
This pull request introduces several improvements and bugfixes around model export, inference, and documentation, with a focus on ONNX and TFLite support. The most significant changes are improved error handling and diagnostics for ambiguous output tensor assignments during inference, an expanded and clarified warning about TFLite export instability, and the addition of ONNX inference helpers for lightweight deployment.

**Export and Inference Improvements**

* [`src/rfdetr/export/_onnx/inference.py`](diffhunk://#diff-f4618ee04efd69b96f9a1707de568b6f85b1af0f0bb3e71b1eb42fa0debb5796R1-R215): Added a new ONNX inference helper module, providing functions for session creation, image preprocessing, and detection decoding, enabling ONNX inference without requiring the full PyTorch stack.
* [`src/rfdetr/export/_tflite/inference.py`](diffhunk://#diff-b97d0660e2c271d5846f16dd227daa7803091b2df11a2bd71cfaace54bd4c820R17-R18): Enhanced TFLite interpreter creation to support `ai_edge_litert` as a backend, and improved output tensor assignment logic to handle ambiguous cases (e.g., when `num_classes==3`) by checking tensor value ranges and swapping assignments if necessary. [[1]](diffhunk://#diff-b97d0660e2c271d5846f16dd227daa7803091b2df11a2bd71cfaace54bd4c820R17-R18) [[2]](diffhunk://#diff-b97d0660e2c271d5846f16dd227daa7803091b2df11a2bd71cfaace54bd4c820L46-R61) [[3]](diffhunk://#diff-b97d0660e2c271d5846f16dd227daa7803091b2df11a2bd71cfaace54bd4c820R190-R207)

**API and Documentation Updates**

[...]

PR #1454, opened 2026-09-11
fix(export): build a real FP16 engine on TensorRT 11

### LINKED ISSUE #1453: build_engine(fp16=True) silently builds an FP32 engine on TensorRT 11
### Search before asking

### Bug

> **Related to #1214, filed deliberately rather than as an oversight.** #1214 reported this same
> breakage and was closed on 2026-07-29 with *"if you still observe these issues, pls feel free to
> reopen it"*. I am not its author, so I cannot reopen it — this is a fresh report taking up that
> invitation, with measurements on an L4 and proof of the engine's actual precision. Glad to have
> this closed as a duplicate if you would rather reopen #1214 instead. Details in *Additional*.

**Executed Colab notebook with all outputs (NVIDIA L4):**
https://colab.research.google.com/drive/1wRDsTbAV800_gFZHzyjP5bUm7tAy7YWO?usp=sharing

### What happens

`pip install rfdetr[tensorrt]` resolves to TensorRT 11 today — the extra declares `tensorrt>=8.6.1`
with no upper bound. On TensorRT 11, `build_engine(..., fp16=True)` logs one warning, **exits 0, and
writes an FP32 engine**. Nothing raises. The engine loads, runs, and can be benchmarked — and the
number is an FP32 number.

Measured on an **NVIDIA L4**, `rfdetr` **1.10.1**, `rfdetr-small` @ 512², batch 1, TensorRT 11.2.1.2.
Engine forward only, 50 warmup + 300 CUDA-event-timed runs:

| Engine | Bytes | % layer I/O tensors in Half | mean ms |
|---|---|---|---|
| `build_engine(fp16=True)` | 119,688,644 | **0.00** | **8.407** |
| same graph cast to FP16, built strongly typed | 65,059,180 | 86.87 | **2.170** |

**3.87× slower** than the FP16 the caller asked for. Anyone benchmarking RF-DETR against other
detectors publishes an FP32 latency labelled FP16.

Build log — the downgrade is visible, but non-fatal:

```
[WARNING] rf-detr - TensorRT 11.2.1.2 does not expose the FP16 builder flag; building an FP32 engine instead. Pass fp16=False to silence this warning.
[INFO]    rf-detr - Building TensorRT engine (fp16=False) from /content/export/rfdetr-small.onnx
    Flags | [STRONGLY_TYPED]
[I] Finished engine building in 19.516 seconds
[I] Saving engine to /content/export/rfdetr-small_fp32.trt     <- exit 0
```

Credit where due: the `_fp32` suffix means the artifact on disk is honestly named. The problem is
that `fp16=True` was requested and silently not delivered.

### Root cause

TensorRT 11 removed weak typing. `BuilderFlag.FP16`, `BuilderFlag.INT8`,
`Builder.platform_has_fast_fp16`, `IInt8Calibrator` and `NetworkDefinitionCreationFlag.EXPLICIT_BATCH`
are all gone. Every network is strongly typed, and **precision is taken from the dtypes in the ONNX
graph**, not from builder flags. Verified on the L4:

```python
tensorrt: 11.2.1.2
BuilderFlag.FP16: False
platform_has_fast_fp16: False
network flags: ['PREFER_AOT_PYTHON_PLUGINS', 'PREFER_JIT_PYTHON_PLUGINS', 'STRONGLY_TYPED']
```

`src/rfdetr/export/_tensorrt.py:132` (line 132 in the released 1.10.1 wheel) probes for the flag and
downgrades when it is absent:

```python
if not hasattr(trt.BuilderFlag, "FP16"):
    logger.warning(
        "TensorRT %s does not expose the FP16 builder flag; building an FP32 engine instead. "
        "Pass fp16=False to silence this warning.", ...)
    fp16 = False
    engine_path = _engine_path(fp16_used=fp16)
```

Per its own comment, that probe exists for **lean/partial TensorRT wheels** that do not expose the
flag. On TensorRT 11 it fires for an entirely different reason — the flag was *deleted from the API* —
and the safe fallback becomes a permanent, silent loss of the requested precision.

The other half of the cause sits upstream of the builder: **the exported ONNX is 100 % float32.**
`export_onnx` performs no cast, and `RFDETR.export`'s `fp16` argument is documented as TensorRT-only.
So under strong typing there is nothing for TensorRT to build in FP16 even in principle. Verified on
the L4: `{'FLOAT': 354}` initializers, zero FLOAT16, and all three graph I/O tensors FLOAT.

Worth noting: simply deleting the guard is *not* the fix. polygraphy then fails loudly instead —
`PolygraphyException: fp16 in CreateConfig is not available on TensorRT version 11.2.1.2`. The guard
is precisely what converts a hard failure into a silent FP32 engine.

### How the FP32-ness is proven

A successful build is not evidence of precision, so this is not argued from the filename or the flag.
TensorRT 11's engine inspector omits the per-layer `Precision` field, so the probe histograms
per-tensor datatypes across every layer's inputs and outputs (this needs
`profiling_verbosity=DETAILED`, which polygraphy sets by default).

The probe is **calibrated in the same notebook, on the same L4**, against two engines of identical
topology that differ only in graph dtype:

| Probe input | reported % Half |
|---|---|
| known-FP32 graph, strongly typed | **0.00** |
| known-FP16 graph, strongly typed | **100.00** |

Applied to what `build_engine(fp16=True)` produced: **0.00 % Half**,
`{'Float': 791, 'Bool': 2, 'Int32': 4}`.

Two independent cross-checks agree — the serialized engine is ~1.84× the size of a real FP16 build,
and the forward is 3.87× slower.

(Exact serialized byte counts drift by <0.05 % between builds because TensorRT picks tactics
non-deterministically; the two figures that do not move are `0.00 %` and `86.87 %`.)

### Environment

Reproduced on the **latest PyPI release**, not just a dev branch:

- **RF-DETR: 1.10.1** (installed by a plain `pip install rfdetr` in Colab on 2026-09-10). The
  `hasattr(trt.BuilderFlag, "FP16")` guard is present at `rfdetr/export/_tensorrt.py:132` in the
  published `rfdetr-1.10.1-py3-none-any.whl`.
- Also present on `develop` @ `75c2503` (1.11.0.dev0), and reported earlier against 1.9.4 in #1214.
- OS / Python: Google Colab, Python 3.13.15
- PyTorch: 2.11.0+cu128 (CUDA 12.8, cuDNN 91900)
- **TensorRT: 11.2.1.2** (`tensorrt_cu13`) — contrast case **10.16.1.11**, where the FP16 flag still
  exists and the engine really is FP16
- GPU: **NVIDIA L4**, compute capability 8.9, 23034 MiB

Second machine, for the same bug on a different model/resolution/CUDA line — RTX 3060 (compute 8.6),
TensorRT 11.2.1.2, `rfdetr-medium` @ 704²: `build_engine(fp16=True)` gives **0.00 % Half** and
26.26 ms; the cast graph gives 88.76 % Half and 5.75 ms.

Exact version line from the notebook:

```
rfdetr 1.10.1 | torch 2.11.0+cu128 | cuda 12.8 | cudnn 91900 | trt 11.2.1.2 | py 3.13.15 | NVIDIA L4
```

Note the TensorRT wheel is the CUDA 13 build (`tensorrt_cu13`) while PyTorch is built against
CUDA 12.8 — TensorRT ships its own CUDA libraries, so this mix is expected and not the cause.

**Executed notebook with outputs (NVIDIA L4):**
https://colab.research.google.com/drive/1wRDsTbAV800_gFZHzyjP5bUm7tAy7YWO?usp=sharing

Self-contained and **needs no dataset** — the bug is in the ONNX→engine precision plumbing and is
data-independent, so the notebook uses stock COCO-pretrained weights and random input tensors. Every
step ends in an `assert`, including a calibration step that verifies the precision probe reports
0.00 % on a known-FP32 engine and 100.00 % on a known-FP16 one before it is trusted on the real one.

Minimal version:

```bash
pip install rfdetr onnx "tensorrt==11.2.1.2" polygraphy
```

```python
import tensorrt as trt
print(trt.__version__, hasattr(trt.BuilderFlag, "FP16"))
# 11.2.1.2 False

from rfdetr import RFDETRSmall
from rfdetr.export._tensorrt import build_engine

onnx_path = str(RFDETRSmall().export(output_dir="/content/export", format="onnx", verbose=False))
engine = build_engine(onnx_path, fp16=True)     # one warning, exit code 0
print(engine)
# /content/export/rfdetr-small_fp32.trt   <- FP32, despite fp16=True
```

To confirm the engine really is FP32 rather than trusting the filename, see section 4 of the
notebook (per-layer datatype histogram via `IEngineInspector`, calibrated in section 4b).

### A real FP16 engine *is* achievable on TensorRT 11

Cast the ONNX graph to FP16 first, then build strongly typed. Two casters were measured; **both
work, and neither hits an unsupported operator** — `GridSample` (deformable attention) converts fine.

| | `onnxconverter-common` | ModelOpt AutoCast (NVIDIA's documented path) |
|---|---|---|
| package size | **428 KB** | ~350 MB of genuinely new deps (cupy, scipy, sympy, pulp) |
| cast time | **2.0 s** | 43 s |
| engine (3060, medium @ 704²) | 72,046,092 B, **88.76 % Half**, 5.75 ms | 72,499,668 B, 87.95 % Half, 5.79 ms |
| IoU vs FP32 (min / mean) | **0.9761** / 0.9957 | 0.9716 / 0.9956 |

`onnxconverter-common` looks like the better trade: its only dependencies (`onnx`, `numpy`,
`protobuf`) are already rfdetr dependencies, so the marginal install cost is effectively zero.

It does need two small fix-ups, both about ten lines, and both found only because the build failed:

1. It does not rewrite the `to` attribute of `Cast` nodes that already exist in the source graph.
   RF-DETR exports 33–35 `Cast(to=FLOAT)` nodes, so the tensor stays float32 while its `value_info`
   claims float16, and TensorRT rejects the graph at the first Conv
   (`INVALID_NODE ... patch_embeddings/projection/Conv`).
2. Its `keep_io_types=True` leaves the FP32 graph input wired straight into an FP16 Conv without
   inserting a Cast, producing the same rejection.

Notably, `onnx.checker.check_model` **passes** on both malformed graphs — only TensorRT's
strongly-typed parser catches them.

**Accuracy cost of FP16** (RTX 3060, 32 real frames, versus ONNX Runtime CPU FP32 on the uncast
graph, boxes matched **by IoU**): 309/309 boxes recovered, 0 unmatched at IoU 0.5, IoU min 0.976 /
mean 0.996; 3/309 flip across a 0.4 score threshold.

> Measurement note for anyone re-running this: comparing `dets[i]` against `ref_dets[i]` *slot-wise*
> reports a 142.9 px deviation. That is FP16 `TopK` reordering the query slots, not box drift. Match
> by IoU before drawing any accuracy conclusion.

### Relation to #1214

#1214 described this correctly and was closed on 2026-07-29 with *"I believe this was resolved in
#853 and #1242 … However, if you still observe these issues, pls feel free to reopen it"*. I am not
the author of #1214 so I cannot reopen it — this is a fresh report taking up that invitation, now
with measurements on an L4. Happy to close this as a duplicate if you would rather reopen #1214.

For the record, on the two PRs cited there: #853 added the polygraphy export path, and #1242 is
`fix(deps): require onnxsim>=0.7.0 …`, unrelated to TensorRT. The `hasattr` FP16 guard actually
arrived in **#1231** (merged 2026-07-24). #1231 fixed a real problem — the hard crash on TensorRT 11 —
so this is not a bad fix; the precision loss it introduced simply went unnoticed.

Two things help explain why it stayed unnoticed:

- **The cookbook can't surface it.** `docs/cookbooks/export-tensorrt.ipynb` installs
  `inference-models[trt10]`, whose extra pins `tensorrt-cu12>=10.0.0,<11.0.0` (PyPI metadata,
  `inference-models` 0.37.0). The cookbook is therefore forced onto TensorRT 10, where FP16 genuinely
  works. It also never asserts precision — it prints the engine size and runs inference, and both
  succeed on an FP32 engine. Its text currently states "The engine is FP16 by default (lowest
  latency); it automatically falls back to FP32 (with a warning) if your TensorRT build doesn't
  expose the FP16 builder flag", which on TensorRT 11 is unconditional.
- **A test asserts the current behaviour is correct.**
  `tests/export/test_tensorrt_export.py::TestBuildEngineFp16Fallback::test_downgrades_to_fp32_when_fp16_flag_unavailable`
  asserts the silent FP32 downgrade, using a fake `__version__ = "11.1.0.106"`. And no test asserts
  that an FP16 request yields an FP16 engine — the GPU e2e test builds with `fp16=False`. From inside
  CI, the behaviour looks correct.

### Separately broken on TensorRT 11: `export/benchmark.py`

`TRTInference` uses six symbols removed in TensorRT 10/11. Four are on the *inference* path, and
since `run_async` is the default (`sync_mode=False`), the class is unusable on TensorRT 11 even for a
pre-built engine:

Line numbers below are on `develop` @ `75c2503`:

| Line | Symbol | Removed in |
|---|---|---|
| 420 | `NetworkDefinitionCreationFlag.EXPLICIT_BATCH` | TensorRT 10 |
| 428 | `BuilderFlag.FP16` | TensorRT 11 |
| 344, 351, 363 | `enumerate(engine)` (`ICudaEngine.__iter__`) | TensorRT 11 |
| 387 | `context.execute_async_v2` | TensorRT 11 |

This one was already broken on TensorRT 10 and is gated behind `[tensorrt-bench]` + `pycuda`, so it
is probably a separate issue — happy to file it if useful.

### Suggested direction

Listing these separately because they are separable, and because two of them are your call rather
than mine:

1. Distinguish "flag absent because lean wheel" from "flag absent because TensorRT ≥ 11 strong
   typing", and never satisfy an `fp16=True` request with an FP32 engine silently.
2. On TensorRT ≥ 11, cast the ONNX and build strongly typed. **This adds a dependency**
   (`onnxconverter-common`) — your call whether that belongs in `[tensorrt]` or a new extra.
3. Raise an actionable error when FP16 is requested and cannot be delivered, naming `tensorrt<11` as
   the fallback. **This is a behaviour change** for anyone currently on TensorRT 11.
4. Replace the test that asserts the downgrade, and add a GPU e2e test that proves FP16 via the
   calibrated probe rather than just checking the file exists.

I have the fix, tests and notebook ready and am glad to open the PR — but points 2 and 3 are
maintainer decisions, so I would rather get a nod on the approach first.

lthphuw:
## Summary

There is no way to build an FP16 TensorRT engine on TensorRT 11. `build_engine(fp16=True)` returns an FP32 one.

TensorRT 11 is strongly typed: `BuilderFlag.FP16` no longer exists and engine precision is taken from the ONNX graph's dtypes instead. The builder-flag path added in #1231 therefore sets nothing, and the existing "no FP16 flag → downgrade to FP32 with a warning" fallback misreads a deliberate API change as a lean wheel. The warning fires and the engine is honestly named `_fp32.trt`, so nothing is *silently* mislabelled — but every TensorRT 11 user loses FP16 entirely, with no way to opt back in.

Measured on an RTX 3060 with `tensorrt==11.3.0.99`, `RFDETRNano` at 384×384:

```
develop      build_engine(onnx, fp16=True)  ->  rfdetr-nano_fp32.trt   111.75 MB
this branch  build_engine(onnx, fp16=True)  ->  rfdetr-nano_fp16.trt    59.82 MB   (1.87x smaller)
```

This builds a real FP16 engine there by casting the ONNX graph to FP16 before the build, keeping FP32 graph inputs and outputs so callers are unaffected. TensorRT < 11 keeps the existing flag path and its lean-wheel warning, unchanged.

## Changes

- **`export/_tensorrt.py`** — on TensorRT >= 11, cast the graph with `onnxconverter-common` and let the builder infer precision. Two fix-ups are needed for the strongly-typed parser to accept the result: pre-existing `Cast(to=FLOAT)` nodes are retargeted to FLOAT16, and explicit boundary casts restore FP32 graph I/O. A missing caster raises instead of silently downgrading. The cast file is a build intermediate and is deleted afterwards, including on a failed build.
- **`export/benchmark.py`** — `TRTInference.build_engine` set `trt.BuilderFlag.FP16` unconditionally, so on TensorRT 11 it raised `AttributeError` and built no engine at all. Same root cause in a worse form; folded into the same commit, happy to split it out.
- **`pyproject.toml`** — `onnxconverter-common` + `onnx` in the `[tensorrt]` extra and the `tests` group.
- **`pyproject.toml` (separate commit)** — a uv override replaces a package's entire requirement set, so the lone 3.12-scoped `ml-dtypes` entry left it uninstalled on every other interpreter. `onnx` requires it unconditionally, so adding `onnx` to the `tests` group made `import onnx` fail collection on 3.10/3.11/3.13. Latent until now: the only jobs installing `onnx` were the GPU job (3.12) and the deps-resolution dry runs, which never import.
- **`detr.py`, `export/main.py`** — the `fp16` docstrings still described the old downgrade-always behaviour.
- **`tests/export/test_tensorrt_export.py`** — CPU-only coverage for all three TensorRT generations plus the artifact cleanup, with the `tensorrt` module pinned per test so the assertions do not flip on a host that really has TensorRT 11.

### On the dependency

`onnxconverter-common` is an 87 KiB pure-Python wheel and adds **no transitive package the extra lacks** (`numpy`/`packaging`/`protobuf` are already there). polygraphy's own `ConvertToFp16` is a thin wrapper over this same library.

I did try the zero-dependency route first, following the ONNX-level rewrite precedent from #1054. A 28-line hand-rolled cast passes `onnx.checker` but fails strict type inference (`Inferred elem type differs from existing elem type: (10) vs (1)`) — exactly what the TensorRT 11 parser rejects. A correct version means reimplementing block-list handling, boundary-cast insertion and subgraph recursion, ~150-250 LOC. #1054 also uses `onnx_graphsurgeon`, which `[tensorrt]` does not ship.

`onnx` itself (~19 MB) is the only real new weight, and it is unavoidable for any graph-level approach: `polygraphy` declares no requirements at all, so `[tensorrt]` never pulled `onnx` in. Happy to move both behind a separate extra if you would rather keep `[tensorrt]` lean.

## Verification

Hardware: RTX 3060 (12 GB), driver 580.173.02. Tested against **both** real TensorRT generations, installed in turn.

```
# TensorRT 10.16.1.11 (weakly typed, BuilderFlag.FP16 present)
pytest tests/export/test_tensorrt_export.py src/rfdetr/export/_tensorrt.py -n 1
# 48 passed, 7 skipped   (includes the GPU end-to-end engine build + PyTorch parity check)
# build_engine(fp16=True) -> 57.89 MB vs 111.23 MB fp32, polygraphy reports Flags | [FP16]

# TensorRT 11.3.0.99 (strongly typed, BuilderFlag.FP16 absent)
pytest tests/export/test_tensorrt_export.py src/rfdetr/export/_tensorrt.py -n 1
# 48 passed, 7 skipped
# build_engine(fp16=True) -> 59.82 MB vs 111.75 MB fp32, polygraphy reports Flags | [STRONGLY_TYPED]
# same script on develop  -> 111.75 MB, ratio 1.00x  (bug reproduced)

# Full CPU suite
pytest src/ tests/ -n 4 -m "not gpu" --ignore=tests/run_smoke_all_models.py \
  --ignore=tests/legacy/test_checkpoint_compat.py --timeout=240
# 4999 passed, 5 failed, 138 skipped, 4 errors in 12:06

pre-commit run --all-files
# 18/19 hooks passed; mypy failed
```

Follow-ups on the two non-green results, both pre-existing:

- 8 of the 9 CPU-suite failures were `-n 4` timeouts under load and pass on a serial re-run. The remaining one, `tests/export/test_onnx_notes.py::_export_tiny_model`, is a doctest that captures a stray `INFO` log line; it fails identically on `develop` in a clean worktree, and this PR touches neither that test nor `_onnx/exporter.py`.
- `mypy` reports 6 errors in `export/_topk.py`, `export/_class_layout.py` and `evaluation/coco_eval.py` — none of which this PR touches. Verified identical on `develop` with this branch's `pyproject.toml` reverted.

Resolution checked per interpreter in a throwaway tree: 3.10 and 3.13 now get `ml-dtypes==0.6.0`, 3.12 still pins `0.5.1` with `numpy==1.26.4` so the onnx2tf/tflite workaround is preserved, and `uv lock` still resolves.

## Performance

Engine size is the verified proxy above. I have not measured inference latency on a controlled setup — if a before/after latency table matters for merge, say so and I will run one.

## Related issues

Fixes #1453.


lthphuw:
@Borda — Copilot's three points are all addressed and verified.

Measured on an RTX 3060, installing each TensorRT in turn:

| TensorRT | `BuilderFlag.FP16` | `fp16=True` → engine | fp32 |
| --- | --- | --- | --- |
| 10.16.1.11 | present | 57.89 MB | 111.23 MB |
| 11.3.0.99 | **absent** | 59.82 MB | 111.75 MB |
| 11.3.0.99 on `develop` | absent | **111.75 MB** | 111.75 MB |

Last row is the bug reproducing. 52 passed against real TRT 11, including the GPU end-to-end build.

On Copilot:

1. The intermediate is now `tempfile.mkstemp` in the source directory — a build can no longer delete a same-named file it did not create. Three regression tests added.
2. Confirmed: TRT 11.3 also removes `NetworkDefinitionCreationFlag.EXPLICIT_BATCH`. Resolved by feature test inside the cleanup scope; `TRTInference.build_engine` now builds a 59.87 MB FP16 engine with nothing leaked.
3. Helper doctest is live, gated with `__doctest_requires__` — a class-level `skipif` does not cover a module-level doctest.

**One commit is outside this PR's scope, deliberately.** `tests/export/test_onnx_notes.py` opens with `importorskip("onnx")`, which the CPU job never satisfies on `develop` — that module has never run in CI. Adding `onnx` to the `tests` group un-skips it and exposes a doctest broken since #1292 (an INFO log lands in doctest's capture buffer). I only redirected stdout in the example; fixing the logger would break `TestGetLoggerUsesRedirectAwareHandlers`. Splitting it out doesn't work either — the test is skipped on `develop`, so the fix couldn't be shown green. Happy to drop it if you'd rather take it separately.

Full CPU suite: `5012 passed, 137 skipped, 0 failed`. `pre-commit` passes everything but `mypy`, whose 6 errors are in files this PR doesn't touch and reproduce on `develop`.

Workflows are at `action_required` — could you approve the run?

lthphuw:
Hi @Borda! I re-verified the latest commit `92e1760e` with real TensorRT, since `88019e5a`/`92e1760e` were written without it. These numbers replace the ones in my earlier comment.

**Setup** (RTX 3060, driver 580; each venv mirrors its CI job):

```bash
# GPU / TensorRT 11: ci-tests-gpu.yml install + the tensorrt extra -> torch 2.10.0+cu130
uv venv --python 3.12
UV_TORCH_BACKEND=auto uv pip install -e ".[onnx,plus,train,augment,visual,data,tensorrt]" --group tests --group ci-gpu-pin
# GPU / TensorRT 10: separate venv, same command + "tensorrt==10.16.1.11"

# CPU: ci-tests-cpu.yml install, Python 3.10 and 3.13 -> torch 2.14.0+cpu
uv venv --python 3.13
uv pip install -e ".[train,augment,cli,visual,data]" --group tests --torch-backend=cpu
```

**`RFDETRNano` 384×384** (latency mean / p50 / p90):

| TensorRT | FP16 engine | FP32 engine |
| --- | --- | --- |
| 11.3.0.99 | 57.6 MiB · 2.28 / 2.20 / 2.56 ms | 107.7 MiB · 7.42 / 7.28 / 7.98 ms |
| 10.16.1.11 | 55.7 MiB · 2.07 / 2.05 / 2.12 ms | 106.7 MiB · 6.92 / 6.79 / 7.40 ms |

- Latency: `TRTInference(sync_mode=True)`, batch 1, 50 warm-up + 500 timed runs, synchronized after each run, host <-> device copies excluded. Not a controlled benchmark; the TRT 11 run shared the machine with other jobs.
- All engines keep FP32 inputs/outputs, and no cast file is left behind. On one random input, FP16 outputs are finite, with a max abs diff vs FP32 of 0.027 (`dets`) and 0.095 (`labels`).

**Tests & checks:**
- `test_tensorrt_export.py` + `_tensorrt.py`/`benchmark.py` doctests: 91 passed, 0 failed on both TRT 11 and TRT 10, including the 4 GPU end-to-end cases. These 4 are marked `e2e_tensorrt`, which `ci-tests-gpu.yml` deselects (`-m "gpu and not e2e_tensorrt"`), so CI never runs them; this local run is their only coverage.
- CPU suite, `ci-tests-cpu.yml` command (without coverage): **5047 passed, 0 failed** on 3.13; **5046 passed, 0 failed** on 3.10.
- `pre-commit` passes (mypy run separately). `mypy` with the `ci-typing.yml` setup has 0 errors, so the 6 errors I mentioned earlier don't reproduce.
- `uv lock --check`, the CI `install-plan` dry-runs and `mkdocs build` all pass. `import onnx` works on 3.10/3.11/3.13.
- Not run locally: the rest of the GPU suite (`-m "gpu and not e2e_tensorrt"`); that one is left to CI.

<details>
<summary>Commands</summary>

```bash
# TensorRT tests (TRT 11 venv, then TRT 10 venv)
pytest tests/export/test_tensorrt_export.py src/rfdetr/export/_tensorrt.py src/rfdetr/export/benchmark.py -n 1 --timeout=600

# CPU suite (CPU venvs, 3.13 and 3.10)
pytest src/ tests/ -n 2 \
  -m "not gpu and not coco17 and not e2e_coreml and not e2e_executorch and not e2e_openvino and not e2e_roboflow and not xla and not tpu" \
  --ignore=tests/run_smoke_all_models.py --ignore=tests/legacy/test_checkpoint_compat.py --timeout=420

# Other checks
SKIP=mypy pre-commit run --all-files
mypy src/rfdetr/ --no-error-summary        # Python 3.10: uv pip install -e ".[train,cli,visual]" --group typing
uvx uv@0.9.26 lock --check
uvx uv@0.9.26 sync --no-default-groups --extra <extra> --python <3.10-3.14> --dry-run
mkdocs build --verbose                     # uv pip install -e ".[plus]" --group docs
```

</details>

Workflows are still at `action_required`. Could you approve the run when you get a chance? Thanks!

Borda (approved):

PR #1455, opened 2026-09-11
fix(training): disable Inductor coalesce tiling

Borda:
- Turn off `torch._inductor.config.triton.coalesce_tiling_analysis` before `torch.compile(..., dynamic=True)`. The analysis is unsupported on the dynamic-shape path but still runs, reaching a size-hint assert in `tiling_utils.get_pw_red_splits` that has no symbolic-shape escape, so whole forward frames fall back to eager.
- Guard the assignment with `hasattr`: the attribute is absent on older torch versions, and assigning an unknown name to the inductor config raises `AttributeError`.

### DIFF
--- a/src/rfdetr/training/module_model.py
+++ b/src/rfdetr/training/module_model.py
@@ -452,16 +452,21 @@ class RFDETRModelModule(LightningModule):
             # tiling_utils.get_pw_red_splits comparing size hints. That assert has no
             # symbolic-shape escape, unlike the CantSplit branch below it, so entire forward
             # frames fall back to eager. Turning the analysis off costs nothing under
-            # dynamic=True. The attribute is absent on older torch versions and assigning an
-            # unknown name to the inductor config raises AttributeError, hence the hasattr guard.
+            # dynamic=True. Passed as a compile option rather than assigned on the inductor
+            # config module, so the default is preserved for any other compilation in this
+            # process. The knob is absent on older torch versions, where passing it would raise
+            # RuntimeError("Unexpected optimization option ..."), hence the hasattr guard.
             # Local import: pulls in inductor, which an uncompiled run never needs.
             import torch._inductor.config as inductor_config
 
+            compile_options: dict[str, Any] = {}
             if hasattr(inductor_config.triton, "coalesce_tiling_analysis"):
-                inductor_config.triton.coalesce_tiling_analysis = False
+                compile_options["triton.coalesce_tiling_analysis"] = False
             # OptimizedModule forwards attribute access to the wrapped LWDETR via
             # __getattr__ at runtime, so self.model keeps working everywhere it's used below.
-            self.model = torch.compile(self.model, dynamic=True)  # type: ignore[assignment]
+            self.model = torch.compile(  # type: ignore[assignment]
+                self.model, dynamic=True, options=compile_options or None
+            )
 
     # ------------------------------------------------------------------
     # PTL lifecycle hooks
--- a/src/rfdetr/training/module_model.py
+++ b/src/rfdetr/training/module_model.py
@@ -446,6 +446,19 @@ class RFDETRModelModule(LightningModule):
             # would cause PendingUnbackedSymbolNotFound (which only occurs without dynamic).
             torch._dynamo.config.suppress_errors = True
             torch._dynamo.config.capture_scalar_outputs = True
+            # Inductor's coalesce tiling analysis is unsupported on the dynamic-shape path
+            # (torch/_inductor/config.py: "coalesce_tiling_analysis does not yet apply to
+            # dynamic shapes"), yet it still runs and reaches an assert in
+            # tiling_utils.get_pw_red_splits comparing size hints. That assert has no
+            # symbolic-shape escape, unlike the CantSplit branch below it, so entire forward
+            # frames fall back to eager. Turning the analysis off costs nothing under
+            # dynamic=True. The attribute is absent on older torch versions and assigning an
+            # unknown name to the inductor config raises AttributeError, hence the hasattr guard.
+            # Local import: pulls in inductor, which an uncompiled run never needs.
+            import torch._inductor.config as inductor_config
+
+            if hasattr(inductor_config.triton, "coalesce_tiling_analysis"):
+                inductor_config.triton.coalesce_tiling_analysis = False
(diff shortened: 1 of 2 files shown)


Borda:
## Title

Inductor: `get_pw_red_splits` raises `AssertionError` under `dynamic=True`, silently dropping compiled frames to eager

## Bug description

`torch.compile(model, dynamic=True)` on CUDA fails to compile whole frames with an `InductorError: AssertionError` raised from `torch/_inductor/tiling_utils.py`, inside the coalesce tiling analysis.

The analysis is documented as not supporting dynamic shapes. From `torch/_inductor/config.py`, present verbatim in both 2.11.0 and 2.14.0:

```python
#  We use a max of 3 if coalesce_tiling_analysis is True, and 2 otherwise.
#  Note - coalesce_tiling_analysis does not yet apply to dynamic shapes.
```

It nevertheless defaults to on and runs regardless of whether shapes are symbolic. In 2.11.0:

```python
coalesce_tiling_analysis: bool = (
    os.environ.get(
        "TORCHINDUCTOR_COALESCE_TILING_ANALYSIS", "1" if not is_fbcode() else "0"
    )
    == "1"
)
```

and the only gate on the call site is that same flag (`codegen/simd.py:1837`):

```python
if torch._inductor.config.triton.coalesce_tiling_analysis:
    if len(nodes) != len(node.get_nodes()):
        assert self.scheduler
        node = scheduler.FusedSchedulerNode(self.scheduler, nodes)
    coalesce_analysis = analyze_memory_coalescing(node)
else:
    coalesce_analysis = None
```

There is no check anywhere on that path for whether the graph has symbolic shapes.

Inside `get_pw_red_splits`, the first branch tries to prove the pointwise product matches `pointwise_numel`:

```python
if n.is_reduction() or V.graph.sizevars.statically_known_equals(
    sympy_product(n._body.sizes[0]), pointwise_numel
):
    return (...)
```

With symbolic sizes this equality is not statically provable, so control falls through to a comparison of *hints* — the example-input values behind the symbols:

```python
assert get_hint(sympy_product(n._body.sizes[0])) == get_hint(
    pointwise_numel * red_numel
)
```

For a fused node built from a split reduction, those hints genuinely differ, and the assert fires.

The asymmetry that makes this look like a missing guard rather than a real invariant violation: further down the same file, `_analyze_memory_coalescing` handles exactly this situation for the split step, and explicitly tolerates dynamic shapes instead of asserting:

```python
except torch._inductor.codegen.simd.CantSplit as e:
    # occasionally with dynamic shapes, we will be unable to prove
    # divisibility
    if not (pointwise_numel.free_symbols or red_numel.free_symbols):
        raise AssertionError(
            "expected dynamic shapes (free symbols) when split fails"
        ) from e
    return None
```

`get_pw_red_splits` has no equivalent escape.

## Impact

With `torch._dynamo.config.suppress_errors = True` this is not a crash — Dynamo logs `WON'T CONVERT` and runs the frame in eager. That makes it worse in practice than a hard failure: compilation silently does nothing for the affected frames, after paying the full compile cost (roughly 50 seconds per frame in the run below), and the only signal is a warning in a very noisy log.

## Reproduction

torch 2.11.0+cu128 on a Colab NVIDIA RTX 6000, Python 3.13. Model is RF-DETR nano (a DETR-family detector), batch size 320, bf16 autocast, `torch.compile(model, dynamic=True)`, `capture_scalar_outputs=True`, `suppress_errors=True`.

Two frames refused, both top-level module forwards. Immediately before the first one, Inductor reports that it split the reduction:

```
torch/_inductor/lowering.py:7836: UserWarning:
Online softmax is disabled on the fly since Inductor decides to
split the reduction.
```

Then:

```
W0911 10:58:38 torch/_dynamo/convert_frame.py:2147] WON'T CONVERT forward .../rfdetr/models/lwdetr.py line 459
W0911 10:58:38 torch/_dynamo/convert_frame.py:2147] due to:
...
  File "torch/_inductor/codegen/simd.py", line 1837, in codegen_node
    coalesce_analysis = analyze_memory_coalescing(node)
  File "torch/_inductor/tiling_utils.py", line 694, in analyze_memory_coalescing
    norm_read_writes = extract_normalized_read_writes(fused_node)
  File "torch/_inductor/tiling_utils.py", line 523, in extract_normalized_read_writes
    pw_splits, red_splits = NodeSplitGetter(node).get_node_splits()
  File "torch/_inductor/tiling_utils.py", line 327, in __init__
    maybe_splits = get_pw_red_splits(
        n, self.pointwise_numel, self.red_numel, none_if_not_divisible=True
    )
  File "torch/_inductor/tiling_utils.py", line 271, in get_pw_red_splits
    assert get_hint(sympy_product(n._body.sizes[0])) == get_hint(
        pointwise_numel * red_numel
    )
torch._inductor.exc.InductorError: AssertionError:
```

A second, identical failure follows at 10:59:28 for another module forward.

The empty message after `AssertionError:` is because the statement is a bare `assert` in 2.11.0. In 2.14.0 it is a `raise AssertionError("expected pointwise sizes to match pointwise_numel * red_numel")` at `tiling_utils.py:310` — same logic, same bug, better diagnostics.

I have not reduced this to a standalone script. If a minimal repro is required before triage, please say so and I will try to build one.

## Workaround

```bash
TORCHINDUCTOR_COALESCE_TILING_ANALYSIS=0
```

or, in code, before the `torch.compile` call:

```python
import torch._inductor.config as inductor_config

inductor_config.triton.coalesce_tiling_analysis = False
```

## Suggested fix

Give `get_pw_red_splits` the same symbolic-shape escape the `CantSplit` handler already has — bail out instead of comparing hints when either numel carries free symbols:

```python
if pointwise_numel.free_symbols or red_numel.free_symbols:
    return None
```

`NodeSplitGetter.__init__` already calls it with `none_if_not_divisible=True` and handles a `None` return, so the call site needs no change. An alternative, if the analysis genuinely has no dynamic-shape support yet as the config comment states, is to skip the analysis entirely when the graph has symbolic shapes, which would also save the wasted analysis work.

## Versions

```
torch          2.11.0+cu128
torchvision    0.26.0+cu128
torchaudio     2.11.0+cu128
```

Python 3.13, CUDA 12.8, NVIDIA RTX 6000, Google Colab.

Confirmed still present by source inspection in torch 2.14.0 (`08187d9e0fba026dc8217405802ab5381dc88d90`), where the assert is at `tiling_utils.py:310`.

Issue #1410, opened 2026-08-30
Investigation: faster training throughput without accuracy regression — starting with `torch.compile`


Borda:
## Motivation

We want lower wall-clock time per epoch at unchanged mAP. `compile=True` exists as an opt-in ([`config.py:502`](https://github.com/roboflow/rf-detr/blob/develop/src/rfdetr/config.py#L502)) and is documented as being "for faster throughput" ([`config.py:467`](https://github.com/roboflow/rf-detr/blob/develop/src/rfdetr/config.py#L467)), but there is no benchmark in the repo establishing what it actually buys, on which hardware, or for which configs. That number is unknown, and it should not be.

This issue is scoped to **measuring and understanding**, not to landing a fix. Any change to defaults or the compile path should follow from numbers produced here.

**Hard constraint throughout: no accuracy regression.** Any candidate change gets validated on mAP, not just ms/step. A throughput win that moves the metric is not a win.

## What we don't know

Open questions, in no particular order. Answering any one of them is useful on its own.

**Baseline**
- What is the steady-state ms/step for eager vs `compile=True`, per model size, on common GPUs (A100, L4, 4090, T4)? With warm-up excluded and validation excluded?
- Does compilation succeed at all under each config we ship? `torch._dynamo.explain(model)(samples)` gives `graph_count` / `graph_break_count` / `break_reasons` without needing a full run.
- How much wall time does the first-step compile cost, and at what epoch count does it pay back?

**Where the time actually goes**
- What fraction of a training step is the model forward/backward, versus the criterion + matcher, the dataloader, GPU-side augmentation, and logging? A `torch.profiler` trace with `schedule(wait=20, warmup=5, active=10)` answers this directly.
- How often is the GPU idle waiting on input? `num_workers` defaults to `2` ([`config.py:1094`](https://github.com/roboflow/rf-detr/blob/develop/src/rfdetr/config.py#L1094)) — is that the binding constraint on typical hardware before anything else matters?
- Where are the host syncs, and how much do they cost? Candidates to look at in a trace: [`matcher.py:726`](https://github.com/roboflow/rf-detr/blob/develop/src/rfdetr/models/matcher.py#L726), [`_assignment.py:83`](https://github.com/roboflow/rf-detr/blob/develop/src/rfdetr/models/_assignment.py#L83).

**Compile configuration**
[...]

PR #1468, opened 2026-09-15
perf(training): add opt-in CUDA graph replay

JESUSROYETH:
## Summary

This adds `ModelConfig.cuda_graphs`, an opt-in CUDA graph path for single-GPU detection training. It captures the model forward and backward once per static input signature, then replays that graph when the same shape returns. The variable-length targets and `SetCriterion` stay eager.

The original detector remains registered on `RFDETRModelModule`; the graph runner is a plain Python helper. This keeps checkpoint keys, optimizer parameters, and EMA ownership unchanged. Capture failures are isolated to one signature and fall back to eager execution. CPU, MPS, distributed, segmentation, keypoint, gradient-checkpointing, and non-BF16 trainer precision (FP16, FP32, FP8) configurations also stay eager with a warning. `cuda_graphs=True` cannot be combined with `compile=True`.

The transformer has one opt-in cache for its immutable `spatial_shapes` tensors. Without it, constructing that CUDA tensor from host shape data inside capture fails. The default eager, compile, and export paths keep their current tensor construction.

## Performance

I measured this through the public PyTorch Lightning training path on one NVIDIA L4: RF-DETR Nano, BF16, batch 4, 384 base resolution, deterministic synthetic detection images and targets, and a fixed 8-resolution multi-scale set (`expanded_scales=False`). This isolates model/criterion/backward/optimizer/framework throughput without dataset decoding or augmentation; I did not run this against a real dataset. `TrainConfig`'s own default, `expanded_scales=True`, resolves to 11 multi-scale resolutions for this model and was not benchmarked; expect proportionally more graph pools and higher peak reserved memory under that default. Each run used 20 untimed batches to cover the shapes, then 30 CUDA-synchronized measured batches per arm. The table reports the median and range over five fresh-process runs after one discarded warmup run.

| Path | Median batch time | Range | Peak allocated | Peak reserved |
| --- | ---: | ---: | ---: | ---: |
| Eager | 149.826 ms | 148.769-152.864 ms | 2,395 MiB | 2,754 MiB |
| CUDA graphs | 101.680 ms | 100.869-102.077 ms | 3,466 MiB | 14,550 MiB |

The median of the five paired per-run reductions was **32.4%** (range 31.4-33.5%). The result was 8/8 captured signatures and zero fallbacks in each run. The cost is memory: peak allocated memory rose by 1,070 MiB, and the caching allocator reserved 5.28x as much memory because the eight graph pools remain alive... so the option stays disabled by default, and the docs ask users to benchmark memory as well as step time.

Short-run numerical differences stayed inside a fresh eager-vs-eager control. The graph arm's median maximum final-parameter difference was 0.005858, versus 0.005925 for the control; its median maximum loss difference was 0.03048, versus 0.04126 for the control. This does not establish full-training accuracy parity, so I am not making that claim here.

Environment: NVIDIA L4 23 GiB, driver 580.178.04, PyTorch 2.9.1+cu129, Python 3.10.12.

## Validation

- `pytest tests/training/test_cuda_graph_step.py tests/training/test_module_model.py::TestCudaGraphLifecycle tests/training/test_module_model.py::TestTrainingStep::test_routes_forward_through_cuda_graph_runner tests/training/test_module_model.py::TestTrainingStep::test_real_model_composes_through_cuda_graph_runner_on_training_step tests/models/test_transformer.py -q`: 100 passed, including real-CUDA capture/replay, accumulated-gradient, and failed-capture fallback tests on an RTX 4060, plus a CPU test that composes a real detector, criterion, `training_step`, backward, and optimizer through the graph-runner wrapper.
- Repository suite: 5,254 passed and 138 skipped, the first environment produced 15 unrelated failures because its subprocesses could not import the src-layout checkout and OpenVINO was already installed where four tests require it to be absent. Exactly those 15 passed in a clean existing reproduction environment with the checkout on `PYTHONPATH`.
- `bin/rfdetr_precommit.sh`: the complete hook set passed, including strict mypy.
- A mutation that cleared gradients during first capture fails the accumulated-gradient regression test; the submitted implementation passes it.

Part of #1410.


Borda:
was just profiling the same but you were faster 🚀

src/rfdetr/config.py
```diff
@@ -500,6 +502,7 @@ class ModelConfig(BaseConfig):
     group_detr: int = 13
     gradient_checkpointing: bool = False
     compile: bool = False
+    cuda_graphs: bool = False
```
Borda, line 524: Update the training COCO 2017 cookbook to have `compile=True` or `cuda_graphs=True` to demonstrate the best performances... so which is better?

src/rfdetr/training/cuda_graph_step.py
```diff
@@ -0,0 +1,148 @@
...
+                    num_warmup_iters=self.num_warmup_iters,
+                    allow_unused_input=True,
+                ),
+            )
+        except Exception:
+            logger.warning(
+                "CUDA graph capture failed for execution signature %s; falling back to eager for this signature.",
+                key,
+                exc_info=True,
+            )
```
Borda: Thinking if we shall escalate this, as User exeception since the combination id not vital and it is given by user, so not fool user that it running just permance did not change

src/rfdetr/config.py
JESUSROYETH: @Borda , Tested both on the current PR head (`85744b74`) using an L4, RF-DETR Nano, BF16, batch 4 and the same 8-scale sequence. In one completed cold run, once compilation settled, `compile=True` measured 138.01 ms/step, while `cuda_graphs=True` measured 101.68 ms/step (median of five runs), so CUDA graphs were 26.3% faster.

The cost is memory: CUDA graphs reserved 14.55 GiB vs 2.35 GiB for compile. Compile also took around 17 minutes to settle across the scales and hit seven AOT fallbacks. So for the best throughput I would use `cuda_graphs=True` in the COCO 2017 cookbook, with a note about the memory cost and its current single-GPU BF16 detection scope. What do you think?

Borda:
running it on Coco2017 and one epoch with rtx6000 takes:
- `compile`: ~7min30s (develop)
- `compile`: ~7min10s (this PR)
- `cuda_graphs`: 8min40s

```
model = RFDETRNano(compile=True | cuda_graphs=True)
model.train(
    dataset_file="coco",
    dataset_dir=COCO_ROOT,
    output_dir=OUTPUT_DIR,
    epochs=EPOCHS,
    batch_size=BATCH_SIZE,
    num_workers=NUM_WORKERS,
    seed=0,
    lr_scheduler="cosine",
    warmup_epochs=1,
    tensorboard=False,
    progress_bar="tqdm",
    amp_dtype="bf16",
    multi_scale=False,
)
```
Am I doing something wrong, or is there some regression?

JESUSROYETH:
Thanks! Which exact RTX 6000 is it, and did you see any Disabling CUDA graphs or capture failed warning? If it is the older Quadro, BF16 falls back to FP16 and the 8m40 run is actually eager. If it is an A6000/Ada and capture worked, then yes, your COCO result shows a regression for this workload.

Borda:
> Thanks! Which exact RTX 6000 is it, and did you see any Disabling CUDA graphs or capture failed warning? If it is the older Quadro, BF16 falls back to FP16 and the 8m40 run is actually eager. If it is an A6000/Ada and capture worked, then yes, your COCO result shows a regression for this workload.

Yeah, it is strange, there was not obvious warning but also the memory usage did not increased as expected so I think all requested fail cuda graph required shall be fatal... 

```
[2026-09-15 14:39:55] [WARNING] rf-detr - Using a different number of positional encodings than DINOv2, which means we're not loading DINOv2 backbone weights. This is not a problem if finetuning a pretrained RF-DETR model.
[2026-09-15 14:39:55] [WARNING] rf-detr - Using patch size 16 instead of 14, which means we're not loading DINOv2 backbone weights. This is not a problem if finetuning a pretrained RF-DETR model.
[2026-09-15 14:39:55] [INFO] rf-detr - File /root/.roboflow/models/rf-detr-nano.pth already exists with correct MD5 hash.
INFO:pytorch_lightning.utilities.rank_zero:Using bfloat16 Automatic Mixed Precision (AMP)
INFO:pytorch_lightning.utilities.rank_zero:GPU available: True (cuda), used: True
INFO:pytorch_lightning.utilities.rank_zero:TPU available: False, using: 0 TPU cores
INFO:pytorch_lightning.utilities.rank_zero:💡 Tip: For seamless cloud logging and experiment tracking, try installing [litlogger](https://pypi.org/project/litlogger/) to enable LitLogger, which logs metrics and artifacts automatically to the Lightning Experiments platform.
[2026-09-15 14:39:56] [INFO] rf-detr - Building COCO train dataset with square resize at resolution 384
[2026-09-15 14:39:56] [INFO] rf-detr - Built 1 Albumentations transforms from config
[2026-09-15 14:39:56] [INFO] rf-detr - Built 1 Albumentations transforms from config
/usr/local/lib/python3.13/dist-packages/lightning_fabric/loggers/csv_logs.py:268: Experiment logs directory /content/output/det_coco2017_nano/ exists and is not empty. Previous log files in this directory will be deleted when the new ones are saved!
loading annotations into memory...
Done (t=4.77s)
creating index...
index created!
[2026-09-15 14:40:04] [INFO] rf-detr - Building COCO val dataset with square resize at resolution 384
loading annotations into memory...
/usr/local/lib/python3.13/dist-packages/pytorch_lightning/callbacks/model_checkpoint.py:881: Checkpoint directory /content/output/det_coco2017_nano exists and is not empty.
INFO:pytorch_lightning.accelerators.cuda:LOCAL_RANK: 0 - CUDA_VISIBLE_DEVICES: [0]
INFO:pytorch_lightning.utilities.rank_zero:Loading `train_dataloader` to estimate number of stepping batches.
Done (t=0.20s)
creating index...
index created!
/usr/local/lib/python3.13/dist-packages/pytorch_lightning/utilities/_pytree.py:21: `isinstance(treespec, LeafSpec)` is deprecated, use `isinstance(treespec, TreeSpec) and treespec.is_leaf()` instead.
/usr/local/lib/python3.13/dist-packages/pytorch_lightning/utilities/model_summary/model_summary.py:242: Precision bf16-mixed is not supported by the model summary.  Estimated model size in MB will not be accurate. Using 32 bits instead.
┏━━━┳━━━━━━━━━━━━━┳━━━━━━━━━━━━━━┳━━━━━━━━┳━━━━━━━┳━━━━━━━┓
┃   ┃ Name        ┃ Type         ┃ Params ┃ Mode  ┃ FLOPs ┃
┡━━━╇━━━━━━━━━━━━━╇━━━━━━━━━━━━━━╇━━━━━━━━╇━━━━━━━╇━━━━━━━┩
│ 0 │ model       │ LWDETR       │ 30.5 M │ train │     0 │
│ 1 │ criterion   │ SetCriterion │      0 │ train │     0 │
│ 2 │ postprocess │ PostProcess  │      0 │ train │     0 │
└───┴─────────────┴──────────────┴────────┴───────┴───────┘
Trainable params: 30.5 M                                                                                           
Non-trainable params: 0                                                                                            
Total params: 30.5 M                                                                                               
Total estimated model params size (MB): 121.869                                                                    
Modules in train mode: 449                                                                                         
Modules in eval mode: 0                                                                                            
Total FLOPs: 0                                                                                                     
INFO:lightning_fabric.utilities.seed:Seed set to 0
Epoch 8:  69%
 1270/1849 [05:59<02:43,  3.54it/s, loss=4.430, loss_cls=0.692, loss_box=0.0374, loss_giou=0.266, train/lr=9.07e-5, val/ema_mAP_50_95=0.461, val/mAP_50_95=0.461, val/F1=0.647, train/loss=4.240, max_mem=40378 MB, free_mem=46586/97250 MB]
/usr/local/lib/python3.13/dist-packages/torch/autograd/graph.py:869: UserWarning: The AccumulateGrad node's stream does not match the stream of the node that produced the incoming gradient. This may incur unnecessary synchronization and break CUDA graph capture if the AccumulateGrad node's stream is the default stream. This mismatch is caused by an AccumulateGrad node created prior to the current iteration being kept alive. This can happen if the autograd graph is still being kept alive by tensors such as the loss, or if you are using DDP, which will stash a reference to the node. To resolve the mismatch, delete all references to the autograd graph or ensure that DDP initialization is performed under the same stream as subsequent forwards. If the mismatch is intentional, you can use torch.autograd.graph.set_warn_on_accumulate_grad_stream_mismatch(False) to suppress this warning. (Triggered internally at /pytorch/torch/csrc/autograd/input_buffer.cpp:240.)
  return Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
[2026-09-15 14:48:47] [INFO] rf-detr - Skipping validation-loss computation: compute_val_loss='auto' found no scheduler or callback monitoring 'val/loss', so no 'val/loss' metric is logged this run. Set compute_val_loss=True to log it every validation epoch (for example to read the curve from metrics.csv, TensorBoard, or Weights & Biases).
            Val (ema) (Epoch 8/40) — Overall Metrics            
┏━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃           mAP            ┃  mAR   ┃         F1 sweep         ┃
┡━━━━━━━━┯━━━━━━━━┯━━━━━━━━╇━━━━━━━━╇━━━━━━━━┯━━━━━━━━┯━━━━━━━━┩
│ 50:95  │   50   │   75   │  @500  │   F1   │  Prec  │ Recall │
├────────┼────────┼────────┼────────┼────────┼────────┼────────┤
│ 0.4612 │ 0.6488 │ 0.4919 │ 0.6414 │ 0.6465 │ 0.7193 │ 0.5957 │
└────────┴────────┴────────┴────────┴────────┴────────┴────────┘
[2026-09-15 14:48:56] [INFO] rf-detr - Best EMA metric improved to 0.4718 (epoch 0)

Borda (dismissed):

Borda (dismissed):

Borda (approved):

PR #1486, opened 2026-09-18
perf(models): build the matcher L1 box cost without `torch.cdist`

tuan3w:
## Summary

`HungarianMatcher` builds its L1 box cost with `torch.cdist(..., p=1)`. `torch.cdist` is a general Minkowski-distance routine, and its CUDA kernel cannot exploit the fact that the feature dimension here is 4, so it becomes the most expensive single operator in the matcher despite doing only four subtractions per pair.

A `torch.profiler` trace of a real `RFDETRMedium` training step (batch 8, resolution 576, `multi_scale`, BF16, person/face dataset with ~22 boxes per image) attributes **37.6 ms/step in 5 calls — 15.3% of all device time** — to `aten::_cdist_forward`. Five calls is one per cost construction: the final decoder layer, three auxiliary layers and the encoder layer.

This replaces it with `rfdetr.utilities.box_ops.pairwise_box_l1_cost`, which broadcasts and reduces over the four box coordinates. That is memory-bound rather than kernel-bound and measured **15-29× faster** on the shapes the matcher produces.

**Results are bit-identical, not approximately equal.** The reduction runs over the feature axis only, so chunking the target axis cannot reassociate anything, and the tests assert `torch.equal` against `torch.cdist` rather than `allclose`. This matters here specifically: the cost matrix feeds a Hungarian solve, and a reassociated sum would change which assignment wins among near-ties. No accuracy change is claimed or possible, because the matrix is unchanged bit for bit.

Two details worth reviewing:

- **Bounded memory.** The naive broadcast would materialise `[*leading, queries, targets, 4]`, which on a crowded batch (`[40, 3900, 300, 4]`) is ~750 MiB. The target axis is therefore processed in chunks sized from `_L1_COST_ELEMENT_BUDGET` (32M elements, 128 MiB), so peak memory stays flat in the batch/query/target product instead of growing with it. At the largest shape measured, peak goes 538 MiB → 612 MiB; at typical shapes the absolute numbers are tens of MiB.
- **Error semantics preserved.** Broadcasting would silently promote a float32/float64 mix, where `torch.cdist` raises. `TestGateRoutesInputsThatDivergeBetweenPaths::test_mixed_dtype_target_boxes_use_full_path` depends on that rejection to keep achieved precision from depending on batch ordering, so mismatched dtypes delegate to `torch.cdist` — an error path, not a hot one. I would rather hear that an explicit `raise` is preferred than guess.

## Changes

- `src/rfdetr/utilities/box_ops.py`: new `pairwise_box_l1_cost` (+ `_L1_COST_ELEMENT_BUDGET`), with a doctest and the chunking rationale in the docstring.
- `src/rfdetr/models/matcher.py`: both cost sites use it — the compact padded path (`_compute_compact_detection_cost_matrix`) and the full-cartesian path (`forward`).
- `tests/utilities/test_box_ops.py`: new `TestPairwiseBoxL1Cost` — `torch.equal` parity against `torch.cdist` for 2-D, batched and forced-chunked inputs (budget monkeypatched so several chunks are required), empty query/target sets, dtype/device preservation, and a `@pytest.mark.gpu` case at the real matcher scale `[40, 3900, 30]`.
- `tests/models/test_matcher.py`: the three spies that counted `torch.cdist` calls now count `pairwise_box_l1_cost` calls. They assert a real invariant — one stacked cost build for all layers versus one per layer, and "stayed on the compact path" versus "fell through to the full path" — which is unchanged; only the operator they hook moved. `expected_cdist_calls` is renamed `expected_cost_builds` accordingly.
- `CHANGELOG.md`: entry under Unreleased → Changed.

## Verification

Run on Ubuntu, Python 3.12, `torch 2.14.0+cu130`, one RTX 5060 Ti.

```
pytest tests/utilities/test_box_ops.py                       # 33 passed
pytest tests/models/ tests/utilities/test_box_ops.py         # 1169 passed, 18 skipped
pytest src/ tests/ -m "not gpu and not coco17 and not integration and not xla and not tpu"
                                                             # 11 failed, 5437 passed, 226 skipped, 25 errors
mypy src/rfdetr/utilities/box_ops.py src/rfdetr/models/matcher.py   # Success: no issues found
pre-commit run <hook> --files <changed files>                # trailing-whitespace, end-of-file-fixer,
                                                             # ruff-check, ruff-format, codespell,
                                                             # docformatter, mdformat: all Passed
```

The GPU-marked parity test ran and passed on the RTX 5060 Ti (it is included in the 1169 above).

**Not verified / caveats, in the interest of not overclaiming:**

- The 11 failures and 25 errors in the full CPU suite are **pre-existing on `develop`**. I confirmed by stashing this branch and re-running the identical command on a clean tree: `11 failed, 5424 passed, 226 skipped, 25 errors`. Same failure set; this branch adds 13 passing tests and no new failures. They are missing `tests/legacy/checkpoints/*.pth` fixtures and optional-dependency collection errors in my environment.
- `pre-commit run --all-files` does **not** complete here: the local `mypy` hook builds its own env and `pydantic-core==2.33.2` fails to compile against the Python 3.14 interpreter pre-commit selects (`maturin`/PyO3 0.24.1 ABI check). That is an environment issue unrelated to this diff, so I ran the individual hooks against the changed files instead, plus `mypy` directly in the project env, as listed above. Please let CI be the authority on the full hook set.
- No TPU/XLA hardware and no A100-class device for a second GPU data point in this run.

## Performance

**Hardware:** NVIDIA RTX 5060 Ti (16 GB, sm_120), `torch 2.14.0+cu130`.

**Operation only** — `triton.testing.do_bench`, shapes taken from a live `RFDETRMedium` run (`[layers × images, queries × group_detr, targets]`), max `|difference|` against `torch.cdist` is exactly `0.0` in every row:

| shape (B, Q, T) | `torch.cdist` | this change | speedup | peak memory |
| --- | --- | --- | --- | --- |
| (8, 3900, 12) | 1.81 ms | 0.07 ms | **27.8×** | 5 → 31 MiB |
| (40, 300, 97) | 4.91 ms | 0.19 ms | **26.0×** | 14 → 42 MiB |
| (40, 3900, 30) | 21.6 ms | 1.13 ms | **19.2×** | 56 → 221 MiB |
| (40, 3900, 97) | 60.5 ms | 4.09 ms | **14.8×** | 176 → 549 MiB |
| (40, 3900, 300) | 195 ms | 13.0 ms | **15.1×** | 538 → 612 MiB |

**End-to-end training**, real Lightning training loop (dataloader → augmentation → AMP forward → matcher/criterion → backward → optimizer → EMA) on a fixed 3,000-image person/face COCO subset, `RFDETRMedium` at resolution 576 with `multi_scale`, batch 8 × `grad_accum_steps` 4, median of 24 measured steps after 8 warm-up steps:

| configuration | before | after | change |
| --- | --- | --- | --- |
| math-SDPA, gradient checkpointing, non-fused AdamW | 12.88 img/s | 14.81 img/s | **+15.0%** |
| fused attention, no checkpointing, fused AdamW | 23.13 img/s | 24.36 img/s | **+5.4%** |

The second row is the more conservative number and the one I would quote: once attention is fused the step is faster overall, so the same absolute saving is a smaller share. Both rows used the SciPy assignment backend and identical data order.

## Related issues

None open that I could find for this. Happy to file one first if you prefer that order for perf changes — I went straight to a draft PR because the diff is small and the numbers are the discussion.

I have two larger follow-ups measured on the same harness, which I did not want to mix into this diff:

1. A fused Triton kernel that builds the **entire** cost matrix (class gather + focal + L1 + box conversion + GIoU + weighted combine) in one launch: 174× faster than the current op chain standalone on `[40, 3900, 30]` (25.4 ms → 0.145 ms), bit-identical on an A100 and within 4e-06 on sm_120, +20% end-to-end. That one adds a Triton dependency and a fallback path, so it deserves its own issue if there is appetite.
2. The SciPy assignment solve currently runs serially over ~520 tiny sub-problems per step (5 layers × 8 images × 13 query groups) while the GPU sits idle — NVTX shows ~102 ms of host time per microbatch inside the criterion. Solving them on a thread pool gave +13.3% end-to-end with bit-identical matchings. That likely belongs in `torch-hungarian` rather than here.


Borda:
Yeah, just created the same :) So happy to continue with yours!

Borda:
> Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](https://cla-assistant.io/roboflow/rf-detr?pullRequest=1486) before we can accept your contribution.
> **1** out of **2** committers have signed the CLA.

@tuan3w ^^ 🙏 or I will proceed with my implementation 🦝

tuan3w:
Hi @Borda ,
Feel free to continue. this is part of my auto-research experiment (https://tuan3w.github.io/rfdetr-training-throughput/).  I will leave it here.
Thanks.

Borda:
> this is part of my auto-research experiment ([tuan3w.github.io/rfdetr-training-throughput](https://tuan3w.github.io/rfdetr-training-throughput/))

Wow, this is great! Would love to learn more about it 🦝

Borda (approved):

Issue #1093, opened 2026-06-06
Multi-GPU DDP training crashes with "unused parameters" error for standard detection models


dohyeonYoon:
### Search before asking

- [x] I have searched the RF-DETR issues and found no similar bug report.

### Bug

Hello Roboflow development team,
I am reaching out to report a bug we recently discovered while using RF-DETR at our company.

### `RuntimeError: unused parameters in DDP` when training standard detection model (no segmentation head) with `strategy='ddp'`

When training a standard RF-DETR detection model (no segmentation head) using multi-GPU DDP,
the following error is raised after the first few training steps.

In my code, if `strategy="ddp_find_unused_parameters_true"` is not passed, the error below occurs:

```python
import os
from rfdetr import RFDETRMedium

model = RFDETRMedium()

model.train(
    dataset_dir="/home/dohyeon/Datasets/dataset_rfdetr_576",
    epochs=1,
    batch_size=4,
    grad_accum_steps=2,  # 2GPU × grad_accum_steps=2 × batch_size=4 = effective 16
    output_dir="output/strategy1",
    progress_bar="tqdm",
    devices=2,
    # strategy="ddp_find_unused_parameters_true",  # NOT set → triggers error
    num_workers=16,
    pin_memory=True,
    resolution=576,
    resume=None,
    checkpoint_interval=1,
    use_ema=True,
    tensorboard=True,
    seed=42,
)
```

```log
(.venv) (base) dohyeon@tidepool-workstation:~/Projects/ClearML/rf-detr$ torchrun --nproc_per_node=2 train_no_aug.py
W0606 14:47:12.778000 97777 torch/distributed/run.py:851]
W0606 14:47:12.778000 97777 torch/distributed/run.py:851] *****************************************
W0606 14:47:12.778000 97777 torch/distributed/run.py:851] Setting OMP_NUM_THREADS environment variable for each process to be 1 in default, to avoid your system being overloaded, please further tune the variable for optimal performance in your application as needed.
W0606 14:47:12.778000 97777 torch/distributed/run.py:851] *****************************************
[2026-06-06 14:47:17] [INFO] rf-detr - File rf-detr-medium.pth already exists with correct MD5 hash.
[2026-06-06 14:47:17] [WARNING] rf-detr - Using a different number of positional encodings than DINOv2, which means we're not loading DINOv2 backbone weights. This is not a problem if finetuning a pretrained RF-DETR model.
[...]

PR #1489, opened 2026-09-18
perf(training): enable DDP static_graph/gradient_as_bucket_view and reuse two-stage selection's class logits

JESUSROYETH:
## What this changes

`trainer.py` sets `DDPStrategy(find_unused_parameters=True)` unconditionally for `strategy='ddp'`/`'auto'`/`'ddp_spawn'`/`'ddp_notebook'`. `find_unused_parameters=True` makes DDP walk the autograd graph after every backward to find which parameters got no gradient this step. `static_graph=True` does the same detection once and then reuses it, per PyTorch's own docs, as long as the set of parameters that can go unused stays fixed across the run.

So I ran an empirical probe: real `RFDETRNano` with `two_stage` both `True` and `False`, and real `RFDETRSegNano` with and without mask targets, forward+backward across several batch/target combinations, including all-empty-target images. That set turns out to be a single, constant parameter across every one of those configurations — `backbone.0.encoder.encoder.embeddings.mask_token`, a DINOv2 masked-pretraining token that ordinary supervised training never touches, not the data-dependent group_detr scenario the existing comment on this flag describes. If the invariant doesn't hold for a configuration outside this probe, PyTorch's own DDP reducer fails the step immediately ("Your training graph has changed in this iteration") instead of silently dropping a gradient, so this is safe to fail loud.

This adds `static_graph=True` and the complementary `gradient_as_bucket_view=True` alongside the existing flag, scoped to what was measured: automatic-optimization models (`grad_accum_steps<=1`, reading the effective value including a caller's `trainer_kwargs` override) that are not keypoint models. Keypoint models and `grad_accum_steps>1` keep the previous behavior unchanged.

Also, `_two_stage_group_selection`'s batched two-stage top-k pass computes every group's classification logits over the full pre-topk encoder memory to rank positions for `torch.topk`, then discards those values. Only the topk *indices* are used, and index selection carries no gradient into the ranked values. `LWDETR.forward` then re-ran `enc_out_class_embed` a second time, per group, on the same positions already selected, to produce the actual `enc_outputs["pred_logits"]`. `enc_out_class_embed` is a plain per-position `nn.Linear`, so gathering its already-computed output at the selected positions gives the exact same result as re-running it there. I checked this against the untouched fallback loop with a new test, at `atol=1e-4, rtol=1e-4` (this file's existing tolerance for batched-GEMM-vs-loop comparisons). This removes that redundant forward+backward entirely on the batched/eligible path (every shipped default). The rare fallback loop (custom/heterogeneous group modules) is untouched.

## Measured effect

2× NVIDIA L4 (driver 580.178.04), torch 2.12.1+cu130, real NCCL, `RFDETRNano`, BF16 autocast, batch 4/GPU, real `SetCriterion`, real `multi_scale` cycling, real `AdamW` step. 8 discarded warm-up steps + 25 measured steps per config, median of 5 independent `torchrun --nproc_per_node=2` runs (fresh NCCL process group each time):

| config | median reduction | range across 5 runs |
|---|---:|---|
| `static_graph`+`gradient_as_bucket_view` alone | 8.70% | 5.06% – 8.93% |
| class-logits reuse alone | 2.75% | 0.78% – 2.82% |
| **both combined** | **10.56%** | **10.37% – 11.41%** |

Neither change alone gives a number that reliably clears an 8% bar on its own — the DDP change dips to 5% in one run, the class-logits reuse alone doesn't clear even 3% in every run. Combined, all 5 independent runs land in a tight 10.37–11.41% band with no overlap below 10%. Correctness was verified under real NCCL: per-parameter gradients came out bit-identical between baseline and candidate across several steps, including an all-empty-target batch. Also, an earlier 2-rank gloo/CPU proof already covered the exact edge case that originally motivated `find_unused_parameters=True` (issue #1093).

## What this does not cover

`world_size>2` and real (non-synthetic) COCO data were not exercised for the throughput number. Keypoint models and `grad_accum_steps>1` are excluded by design, not by a testing gap: `static_graph` stays `False` there, matching current behavior exactly. No mAP/convergence comparison over a real training run. `gradient_as_bucket_view`'s interaction with EMA is reasoned safe (EMA reads `.detach()`'d parameter data, not `.grad`), but that was not directly measured.

## Tests

`test_two_stage_group_selection_class_logits_reuse_matches_recomputed_loop` (new, `tests/models/test_transformer.py`) builds a real production `RFDETRNano` two ways — the new default path and the untouched fallback loop forced via a monkeypatch — and checks that `enc_outputs["pred_logits"]` matches within tolerance, and that `enc_out_class_embed`'s parameters receive gradient through the reused value. That gradient has to match the fallback loop's independently computed gradient within the same tolerance, a finite gradient alone isn't enough. `test_two_stage_group_selection_bf16_produces_finite_valid_selection_with_gradients` (existing, extended) now also captures the reused class logits under `bf16` autocast — the dtype the measurement above actually runs under — and asserts they are finite and that `enc_out_class_embed` receives real gradient through them.

`TestBuildTrainerDDPStaticGraph` (new, `tests/training/test_build_trainer.py`, 7 cases) covers: detection and segmentation both enable `static_graph`/`gradient_as_bucket_view`; `strategy='auto'`+`devices>1` and `strategy='ddp_spawn'` both enable it through the same paths; `grad_accum_steps>1` disables it; keypoint models disable it; and a caller passing `accumulate_grad_batches=` directly (bypassing `tc.grad_accum_steps`) also disables it. This last case is a regression test for a real crash (`RuntimeError: expect_autograd_hooks_ INTERNAL ASSERT FAILED` in PyTorch's `reducer.cpp`): an existing test in this repo's suite, `test_ddp_spawn_preserves_minimum_optimizer_steps[2]` (a real 2-rank gloo run), caught that exact gap in the original gating logic before this PR's own new tests did.

Full `tests/models/` + `tests/training/` suite (excluding one unrelated pre-existing failure cluster, `TestCudaGraphLifecycle`, confirmed failing identically on unmodified `develop`, and the always-skipped smoke suite): 2530 passed, 0 failed. `pre-commit run --all-files` clean on every touched file (the only remaining `mypy` errors are pre-existing and untouched by this diff: `coco_eval.py:178/190`, `weights.py:691`, `detr.py:91`).

Related to #1410.


src/rfdetr/models/lwdetr.py
```diff
@@ -511,11 +511,15 @@ def forward(self, samples: NestedTensor, targets: list[dict[str, Tensor]] | None
             cross_attn_srcs=cross_attn_srcs,
         )
         if self.use_grouppose_keypoints:
-            hs, ref_unsigmoid, hs_enc, ref_enc, keypoint_hs, enc_kp_predictions, _ = transformer_outputs
+            hs, ref_unsigmoid, hs_enc, ref_enc, keypoint_hs, enc_kp_predictions, _ = transformer_outputs[:7]
```
Borda:
```suggestion
            hs, ref_unsigmoid, hs_enc, ref_enc, keypoint_hs, enc_kp_predictions = transformer_outputs[:6]
```

CHANGELOG.md
```diff
@@ -16,6 +16,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
...
...ation) and `grad_accum_steps>1` (DDP's `no_sync()` across multiple backward calls per optimizer step) keep the previous behavior (`static_graph=False`), since neither was exercised by the measurement below. On 2x NVIDIA L4 (driver 580.178.04, torch 2.12.1+cu130, real NCCL, RF-DETR Nano, BF16, batch 4/GPU, real `SetCriterion`, real `multi_scale` cycling, real AdamW step, 25 measured steps after 8 discarded warm-ups, median of 5 independent process-group runs): `static_graph=True`+`gradient_as_bucket_view=True` alone gave a median **-8.70%** full training step (range -5.06% to -8.93%); combined with reusing the two-stage selection's already-computed classification logits instead of recomputing them (see the next entry) gave a median **-10.56%** (range -10.37% to -11.41%, tight and consistent across all 5 runs), zero errors in any run. Gradient correctness was verified under real NCCL (per-parameter gradients bit-identical between baseline and candidate across several steps including an all-empty-target batch) in addition to an earlier 2-rank gloo/CPU proof. `world_size>2` and real (non-synthetic) COCO data were not exercised. ([#1410](https://github.com/roboflow/rf-detr/issues/1410))
```
Borda:
```suggestion
- DDP training (`strategy='ddp'`/`'auto'`/`'ddp_spawn'`/`'ddp_notebook'`) now also sets `static_graph=True` and `gradient_as_bucket_view=True` on `DDPStrategy`, alongside the existing unconditional `find_unused_parameters=True`. PyTorch's own DDP documents `static_graph=True` as skipping the per-iteration autograd-graph search `find_unused_parameters=True` otherwise repeats every step, as long as the set of parameters that can go unused across the run stays fixed; an empirical probe (real `RFDETRNano`/`RFDETRSegNano` forward+backward, several batch/target combinations including all-empty-target images) found that set is a single, constant parameter, `backbone.0.encoder.encoder.embeddings.mask_token` (a DINOv2 masked-pretraining token untouched in ordinary supervised training), not the data-dependent set the comment on this flag used to describe — so the precondition holds. `gradient_as_bucket_view=True` is bundled in as a complementary, lower-risk flag measured alongside it. Scope is limited to what was measured: automatic-optimization models (detection, segmentation) with `grad_accum_steps<=1` (reading the effective `accumulate_grad_batches`, including a caller's `trainer_kwargs` override); keypoint models (manual optimization) and `grad_accum_steps>1` (DDP's `no_sync()` across multiple backward calls per optimizer step) keep the previous behavior (`static_graph=False`), since neither was exercised by the measurement below. On 2x NVIDIA L4 (driver 580.178.04, torch 2.12.1+cu130, real NCCL, RF-DETR Nano, BF16, batch 4/GPU, real `SetCriterion`, real `multi_scale` cycling, real AdamW step, 25 measured steps after 8 discarded warm-ups, median of 5 independent process-group runs): `static_graph=True`+`gradient_as_bucket_view=True` alone gave a median **-8.70%** full training step (range -5.06% to -8.93%); combined with reusing the two-stage selection's already-computed classification logits instead of recomputing them (see the next entry) gave a median **-10.56%** (range -10.37% to -11.41%, tight and consistent across all 5 runs), zero errors in any run. Gradient correctness was verified under real NCCL (per-parameter gradients bit-identical between baseline and candidate across several steps including an all-empty-target batch) in addition to an earlier 2-rank gloo/CPU proof. `world_size>2` and real (non-synthetic) COCO data were not exercised. ([#1489](https://github.com/roboflow/rf-detr/issues/1489))
```

CHANGELOG.md
```diff
@@ -16,6 +16,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
...
...`enc_out_class_embed` a second time, per group, on the same gathered hidden state to produce `enc_outputs["pred_logits"]`. `enc_out_class_embed` is a plain per-position `nn.Linear`, so gathering its output is what re-running it on the same gathered subset would produce, within the same float32 GEMM-accumulation-order tolerance already documented for the sibling batched selection; a new test (`test_two_stage_group_selection_class_logits_reuse_matches_recomputed_loop`) builds a real production `RFDETRNano` and checks this against the untouched fallback loop directly. Only the fast path (eligible `group_detr>1` configs, i.e. every shipped default) is changed; the rare fallback loop (custom/heterogeneous group modules) is untouched and still recomputes. On its own this removed an entire redundant per-group forward+backward and measured a median **-2.75%** full training step on 2x L4 (range -0.78% to -2.82% across 5 runs) — below this project's 3% GPU-training bar alone — but composes with the `static_graph`/`gradient_as_bucket_view` change above to a reliable **-10.56%** combined median; see that entry for the full measurement. ([#1410](https://github.com/roboflow/rf-detr/issues/1410))
```
Borda:
```suggestion
- `Transformer._two_stage_group_selection`'s batched two-stage top-k pass now also gathers `enc_out_class_embed`'s already-computed per-position output at the selected positions and returns it, instead of the caller (`LWDETR.forward`) re-running `enc_out_class_embed` a second time, per group, on the same gathered hidden state to produce `enc_outputs["pred_logits"]`. `enc_out_class_embed` is a plain per-position `nn.Linear`, so gathering its output is what re-running it on the same gathered subset would produce, within the same float32 GEMM-accumulation-order tolerance already documented for the sibling batched selection; a new test (`test_two_stage_group_selection_class_logits_reuse_matches_recomputed_loop`) builds a real production `RFDETRNano` and checks this against the untouched fallback loop directly. Only the fast path (eligible `group_detr>1` configs, i.e. every shipped default) is changed; the rare fallback loop (custom/heterogeneous group modules) is untouched and still recomputes. On its own this removed an entire redundant per-group forward+backward and measured a median **-2.75%** full training step on 2x L4 (range -0.78% to -2.82% across 5 runs) — below this project's 3% GPU-training bar alone — but composes with the `static_graph`/`gradient_as_bucket_view` change above to a reliable **-10.56%** combined median; see that entry for the full measurement. ([#1489](https://github.com/roboflow/rf-detr/issues/1489))
```

Borda (dismissed):

Borda (approved):

PR #1491, opened 2026-09-19
perf(training): add "vernier" as a fourth COCO eval backend

NoeFontana:
## Summary

Adds `"vernier"` as a fourth value of `TrainConfig.eval_backend`. Vernier is faster than alternative; both single-threaded and multi-threaded. https://noefontana.github.io/vernier/0.5/benchmarks/

The default stays `"hotcoco"` as we may want to A/B test performance in a training run before changing the default but I am happy to switch now if desired. 

Keypoint OKS evaluation and the ONNX/TensorRT benchmark evaluator in `rfdetr.evaluation.coco_eval` are untouched.

## Changes

- **`_VernierBackend` + registry entry.** `eval_backend="vernier"`. TorchCocoMetrics adapters.
- Vernier num_threads is set by `torch.get_num_threads()`.
- Mask-only evaluation is rejected at construction rather than at `compute()`. rf-detr always provide boxes with masks so it's not an issue.
- minor refactoring in coco_map so logic is reused between hotcoco and vernier backend integration
- parametrize common tests for the different eval backends as to limit divergence and duplication

## Performance

**Hardware.** AMD EPYC-Milan, 8 vCPU (4 physical cores × 2 SMT), 32 GB, Linux 6.8, Python 3.12.

**Method.**
- Only `compute()` is timed — `update()` and the forward pass are backend-independent.
- Every backend's output is asserted bit-equal to `faster_coco_eval`'s before any timing is reported.
- `_computed` is cleared before each call, since `compute()` is memoized.
- **Median of 9 repetitions**, with the interquartile range as a percentage of the median in brackets.
- Cores are pinned with `taskset`, which no thread pool can escape. 

---

**On Real Data**

500 images from COCO val 2017.
Predictions generated by RFDTRNano (150k detections) and RFDETRSegNano (45.9k detections with mask) using low thresholds (resp. 0.001 and 0.02)

  **bbox — 500 images, 150k detections, 3,564 ground truths** (mAP 0.5267, mAP@50 0.7017)

  | Backend | median | IQR | vs `faster_coco_eval` |
  | --- | --- | --- | --- |
  | `vernier` | **0.081 s** | 3.1% | 18.8x |
  | `hotcoco` (default) | 0.180 s | 1.4% | 8.4x |
  | `ufcoco` | 0.801 s | 7.5% | 1.9x |
  | `faster_coco_eval` | 1.519 s | 19.6% | 1.0x |

  **bbox + segm — 500 images, 45.9k detections, 3,525 ground-truth masks** (bbox mAP 0.5280, segm mAP 0.4411)

  | Backend | median | IQR | vs `faster_coco_eval` |
  | --- | --- | --- | --- |
  | `vernier` | **0.168 s** | 6.6% | 14.7x |
  | `ufcoco` | 0.621 s | 45.0% | 4.0x |
  | `hotcoco` (default) | 1.206 s | 24.0% | 2.0x |
  | `faster_coco_eval` | 2.458 s | 12.0% | 1.0x |

  Against the hotcoco, vernier is **2.2x on bbox** and **5.6-7.3x on bbox+segm**.

---

**On Synthetic data**

**bbox — 1,000 images × 300 detections, 20 classes, `eval_max_dets=500`** (seconds)

| Backend | 1 core | 2 cores | 4 cores | 8 vCPUs (SMT) |
| --- | --- | --- | --- | --- |
| `vernier` | **0.320** [4.5%] | **0.222** [1.5%] | **0.190** [1.9%] | **0.187** [3.5%] |
| `hotcoco` (default) | 0.701 [5.9%] | 0.511 [2.3%] | 0.416 [2.5%] | 0.404 [6.7%] |
| `ufcoco` | 1.388 [0.9%] | 1.319 [1.2%] | 1.283 [1.4%] | 1.305 [0.9%] |
| `faster_coco_eval` | 2.536 [12.7%] | 2.210 [13.5%] | 2.084 [15.6%] | 2.097 [14.4%] |

**bbox + segm — 1,000 images × 100 detections, 20 classes, 256 px masks** (seconds)

| Backend | 1 core | 2 cores | 4 cores | 8 vCPUs (SMT) |
| --- | --- | --- | --- | --- |
| `vernier` | **0.382** [3.5%] | **0.296** [1.7%] | **0.234** [4.2%] | **0.225** [2.2%] |
| `ufcoco` | 1.030 [9.3%] | 0.931 [3.5%] | 0.932 [2.8%] | 0.913 [1.9%] |
| `hotcoco` (default) | 2.270 [10.9%] | 2.079 [6.9%] | 1.976 [5.9%] | 1.936 [6.1%] |
| `faster_coco_eval` | 4.843 [4.6%] | 4.821 [5.0%] | 4.730 [5.2%] | 4.707 [0.8%] |

**COCO-val-shaped — 5,000 images × 300 detections, 80 classes, bbox, 4 cores** (seconds)

| `vernier` | `hotcoco` | `ufcoco` | `faster_coco_eval` |
| --- | --- | --- | --- |
| **1.860** [2.8%] | 3.182 [9.0%] | 4.964 [12.1%] | 11.124 [9.1%] |

**TLDR**
- vernier is **2.1–2.3× `hotcoco`** on bbox at every core count and **5.9–8.6×** on bbox+segm; the margin narrows to 1.7× at COCO-val's 80 classes.
- vernier and `hotcoco` scale alike (1.68× and 1.69× from one core to four); `ufcoco` (1.08×) and `faster_coco_eval` (1.22×) barely scale, so their gap widens with cores rather than closing.
- IQRs sit at or under 7% everywhere except `faster_coco_eval` on bbox (13–16%), which is the least stable number in these tables — the reason for medians of 9 rather than single calls. On a smaller 200-image bbox config `ufcoco` was outright bimodal (modes at 0.086 s and 0.344 s, IQR 77% of the median), so its single-call numbers are not trustworthy.

## Verification

```
uv run --no-sync pytest src/ tests/ scripts/ -n 4 -m "not gpu" \
  --ignore=tests/run_smoke_all_models.py --ignore=tests/legacy/test_checkpoint_compat.py \
  --timeout=420             # 5666 passed, 171 skipped in 403.95s
pre-commit run --all-files  # passed except three mypy errors that reproduce on develop unchanged
                            # (src/rfdetr/evaluation/coco_eval.py:178,190 and src/rfdetr/detr.py:91 —
                            #  files this PR does not touch)
```

Correctness is pinned at two levels rather than one:

- **Same document.** `CocoDataset.dataset_hash` proves the columnar ground truth is byte-identical to the COCO-JSON route built from TorchMetrics' own `_get_coco_format`, across six fixtures. Most differ only in ways AP cannot see — an image entry with no annotations, a category name, an `area` on an annotation that matches nothing — so a metrics-only check would pass while the document diverged.
- **Same numbers.** Exact equality (`rtol=0, atol=0`) against `faster_coco_eval` over tied scores, a crowd target with a supplied area, images missing either side, both IoU types, and `eval_max_dets` at and above pycocotools' fixed 100.
- **Mutation-checked.** Each of these breaks the suite rather than passing quietly (failing vernier tests in brackets): dropping the per-element `area` fallback [6], starting annotation IDs at 0 [6], narrowing `iscrowd` to `uint8` so 256 wraps to 0 [1], computing a mask annotation's area from its box [1], leaving every image size at `0x0` [7].
- `test_vernier_parity_modes_differ_only_on_the_l9_aggregate` turns the claim that only quirk L9 reaches this path into a test, so a future vernier minor that adds a correction reaching RF-DETR fails loudly.

Not verified: no GPU available here, so `-m gpu` was not run; this path is CPU-side `compute()` and carries no GPU-specific branch.


docs/learn/train/training-parameters.md
```diff
@@ -325,58 +326,112 @@ With `pad_targets_to` set, only the training loader pads; validation, test and p
...
+| `prefetch_factor`            | int                                                        | None           | Number of batches prefetched per worker. None uses PyTorch default.                                                                                                                                                       |
+| `pack_targets`               | bool                                                       | True           | Concatenate target dicts before crossing the DataLoader worker boundary. See DataLoader Tuning; set False to opt out.                                                                                                     |
+| `pad_targets_to`             | `int \| None`                                              | `None`         | Pad training targets to this many rows, for a shape-stable XLA/TPU graph. See DataLoader Tuning; training loader only.                                                                                                    |
```
Borda, line 437: seem duplictaed

NoeFontana:
Now also benchmarked against RFDetrNano and RefDETRSegNano predictions on a COCOVal2017 subset, showing vernier is 2.2x and ~6x faster than hotcoco on these workloads

docs/learn/train/training-parameters.md
NoeFontana: Addressed in https://github.com/roboflow/rf-detr/pull/1491/commits/082dfb1777aad958ce21747833839f59f02cf090

docs/learn/train/training-parameters.md
```diff
@@ -325,58 +326,112 @@ With `pad_targets_to` set, only the training loader pads; validation, test and p
...
+|                              |                                                            |                |                                                                                                                                                                                                                           |
+| Parameter                    | Type                                                       | Default        | Description                                                                                                                                                                                                               |
```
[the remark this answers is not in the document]
NoeFontana: Addressed in https://github.com/roboflow/rf-detr/commit/082dfb1777aad958ce21747833839f59f02cf090

tests/training/test_coco_map.py
```diff
@@ -1075,6 +1097,273 @@ def test_ufcoco_backend_picks_up_the_optional_package_and_survives_pickling() ->
...
+        "sync_on_compute": False,
+    }
+
+    def results(parity_mode: str) -> dict[str, torch.Tensor]:
+        with patch("rfdetr.training.coco_map._VERNIER_PARITY_MODE", parity_mode):
+            metric = OnePassCocoMeanAveragePrecision(**kwargs)
+            metric.update(copy.deepcopy(predictions), copy.deepcopy(targets))
+            return metric.compute()
+
+    strict, corrected = results("strict"), results("corrected")
```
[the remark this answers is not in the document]
NoeFontana: A similar pattern already pre-existed in this file.

I think it may be more readable and maintainable to keep the helper but I don't mind removing it if preferred.

src/rfdetr/training/coco_map.py
```diff
@@ -502,11 +585,247 @@ def compute(self) -> dict[str, Tensor]:
                     coco_eval.stats, prefix=prefix, max_detection_thresholds=self.max_detection_thresholds
                 )
             )
-            result.update(self._reduce_per_class(coco_eval, prefix, classes))
+            result.update(self._reduce_per_class(getattr(coco_eval, "eval", None), prefix, classes))
 
         result["classes"] = torch.tensor(classes, dtype=torch.int32)
         return result
 
+    def _vernier_results(self, classes: list[int]) -> dict[str, Tensor]:
```
Borda, line 593: @NoeFontana this works, but do you think it could be upstreamed into to `vernier` since we use Troecmetrics as orchestrator so I believe it would be useful for more aplications/addoption

Borda (approved):

src/rfdetr/training/coco_map.py
NoeFontana: @Borda I can take a look at it. I would likely leave it behind an extra dependency flag.

To clarify, would you like to see a torch integration (returning the same type as the vernier_results helper) or a deeper torchmetrics integration?

Borda: probably a bit of both, which does not mean you shall abandon your API but rather expand in a healthy way, so it is not binding only RFDETR but any simolar usecase

Borda: >  I would likely leave it behind an extra dependency flag.

@NoeFontana, so you suggest not using it as the default?

NoeFontana: I meant that, within vernier, I would leave integrations to libraries (torch/torchmetrics) as project.optional-dependencies as to not add large packages that aren't needed for an evaluation engine.

It would be exposed in rf-detr and other training frameworks as vernier[torchmetrics]

Borda: You don't need to add heavy dependences, just enable to be simply plugged in...

NoeFontana: https://github.com/roboflow/rf-detr/pull/1512

PR #1492, opened 2026-09-19
perf(models): keep the solved assignment indices on the device

tuan3w:
## Summary

`_solve_to_indices` ends with `pair = torch.stack((rows, cols)).cpu()`, documented as "the single device-to-host transfer for the whole step". The batched solver produced that assignment on the device, and every consumer uses the pairs to index CUDA prediction and target tensors — so the one deliberate D2H transfer is paid back as an implicit H2D copy at each of those gathers.

Measured on the real training loop (RF-DETR Medium, batch 8, CUDA, `torch.cuda.set_sync_debug_mode("warn")`, warnings attributed to the innermost frame), `develop` synchronises **164.3 times per step**, and the top sites are all consumers of these indices:

| syncs/step | site (on `develop`) |
|---:|---|
| 38.3 | `criterion.py:962` `_get_matched_targets()` |
| 38.3 | `criterion.py:964` `_get_matched_targets()` |
| 20.0 | `autograd/graph.py:1072` `_engine_run_backward()` |
| 10.0 each | `criterion.py:580`, `:594`, `:598` `loss_labels()` |

Keeping the pairs on the device they were solved on takes that to **14.7 per step** (−91%), reproducible to the decimal across runs. What remains is `loss_cardinality` (5.0), Lightning's `to_item` (3.0) and two inside `torch_linear_assignment` itself — all out of scope here.

This is a **host-synchronisation fix, not a throughput claim**: see Performance below, where the img/s effect on my hardware is inside the noise floor. It also unblocks the obvious follow-up — a solver whose result never leaves the device is only worth having if the unpack does not drag it back.

## Changes

- `_solve_to_indices` returns the index pairs on the solver's device instead of copying them to the host.
- `_assign_padded` drops the padded columns without a boolean mask, whose kept count only the device knows. A **stable** sort on the padded flag moves padded columns to the end of each problem while leaving the real ones in the ascending row order boolean indexing produced, so the output is unchanged; the positions to read then follow from `sizes`, which the caller already passed. Added `_real_pair_positions` for that index, built on the host and transferred once.
- `size_per_problem` is built on the host and moved with `non_blocking=True`; constructing it directly on the device copies from pageable memory and blocks.
- `assign_many_bucketed`'s all-empty early return now places its empty tensors on the input device instead of hardcoding CPU, which previously handed the criterion host indices for a device batch.
- Docstrings updated: the module's "single device-to-host transfer" note, and the three `Returns` sections that promised CPU tensors.
- `_assert_same_indices` compares on the host, because the cross-backend parity tests hand it a CUDA-solved assignment and its SciPy-solved reference, and `torch.equal` rejects tensors on different devices.
- Two GPU tests that asserted `"assignment indices must return on CPU"` now assert the indices stay on the solving device — that contract is what this PR changes.

## Verification

```bash
# targeted, including the new tests
uv run --no-sync pytest tests/models/test_matcher.py -k "AssignmentBucketing or AssignmentOnCUDA or MasksHybridPathOnCUDA or CompactPathOnCUDA"
# 15 passed

# full CPU suite, as ci-tests-cpu.yml runs it
uv run --no-sync pytest src/ tests/ scripts/ -n 2 \
  -m "not gpu and not coco17 and not integration and not xla and not tpu" \
  --ignore=tests/run_smoke_all_models.py --ignore=tests/legacy/test_checkpoint_compat.py --timeout=240
# 5625 passed, 120 skipped, 5 failed

pre-commit run --files src/rfdetr/models/_assignment.py tests/models/test_matcher.py
# all hooks Passed (ruff, ruff-format, docformatter, codespell, insert-license, ...)

uv run --group typing --extra train --extra cli --extra visual mypy src/rfdetr/models/_assignment.py
# clean
```

The 5 CPU failures are all `tests/training/test_module_model.py::TestCudaGraphLifecycle` with `TypeError: dynamic spec expects a dict, ShapesSpec, or ParamsSpec, got MagicMock`. **They are pre-existing**: I reverted both of my files onto `5b39cbc` and got the same 5 failures, so they are a torch-2.14 interaction with those tests' `MagicMock`, unrelated to this change.

New tests, written before the fix and confirmed to fail without it:

- `TestGpuAssignmentOnCUDA::test_indices_stay_on_cuda` — fails on `develop` (`assert False`).
- `TestGpuAssignmentOnCUDA::test_unpacking_the_solve_does_not_synchronize` — fails on `develop` naming the exact line: `_assignment.py:83 in _solve_to_indices`. It attributes warnings to the **innermost** frame, because this module also sits on the stack of the dependency's own synchronising calls; it skips both `warnings.py` (≤3.13) and `_py_warnings.py` (3.14+), and missing that second name makes the assertion silently vacuous.
- `TestGpuAssignmentBucketing::test_indices_come_back_on_the_cost_matrix_device` — CPU, parametrized over a populated and an all-empty batch; the empty case covers the hardcoded-CPU early return.

Equivalence, beyond the suite: `assign_many_bucketed` returns byte-identical indices to the SciPy per-layer reference across `sizes`/`group_detr`/layer-count combinations including empty images and `group_detr=13` (`torch.equal` on every pair), on CPU and CUDA.

Not verified: no mAP run. This does not change any assignment — the sort is stable and the output order is the same — but that is an argument from the code plus index-level equality, not a training result.

## Performance

**No throughput claim: on my hardware the effect is below the noise floor, and I would rather say so than dress it up.**

Paired alternating runs, two git worktrees selected by `PYTHONPATH` (RTX 5060 Ti, 24 SMs, torch 2.14.0+cu130, Medium, batch 8, fixed 576px so multi-scale reshaping does not add variance):

| config | paired deltas (after vs before) | median |
|---|---|---|
| eager, 40 steps, 5 rounds | +2.27%, −0.68%, +3.86%, −9.08%, +3.01% | +2.27% |
| **compiled**, 40–60 steps, 9 rounds | +1.80%, +1.83%, −6.55%, +0.85%, +1.94%, −0.44%, −0.42%, −12.49%, +0.23% | **+0.23%** |

Five of nine compiled rounds are positive, but two rounds land at −6.6% and −12.5%, and re-measuring *identical* code on this box spreads by a comparable amount. So these runs cannot resolve a ~1% effect.

That is also the expected outcome here rather than a surprise: with Nsight this configuration sits at **89–90% device-busy**, so the host is already running ahead of the GPU and removing its stalls has nothing to give back. The machines in #1410 that are host-bound — the L4 at 19–45% GPU utilisation, issuing 4,200–4,800 kernels per step — are where 150 fewer synchronisations per step should actually convert into wall time. I do not have one of those, so the sync count above is the evidence I can stand behind, and I would welcome a re-measurement on that hardware before anyone treats this as a speedup.

## Related issues

Part of #1410 (throughput investigation). Follows #1486.

---

*Context: this came out of an autoresearch session driven by the [Oh My Pi](https://tuan3w.github.io/rfdetr-training-throughput/) coding agent (Claude Opus), which built the measurement harness, ran the profiling and drafted this change. Full method, every kept and rejected experiment, and the raw session log: https://tuan3w.github.io/rfdetr-training-throughput/ — I reviewed the diff, re-ran the verification above myself, and can answer follow-ups.*


tuan3w:
Closing this myself: I could not demonstrate a throughput improvement, and that is the bar this repo holds perf changes to.

What the change does is measurable and deterministic — host synchronisations per training step drop from 164.3 to 14.7, with `develop`'s top offenders being the consumers of these very indices (`criterion.py:962`/`:964` at 38.3/step each). What I could not show is that it converts into wall time: across 9 paired alternating rounds under `torch.compile` the median was +0.23%, with two rounds at −6.6% and −12.5%, and identical code re-measures with comparable spread on my box. At 89–90% device-busy this GPU is GPU-bound, so the host already runs ahead and removing its stalls has nothing to give back.

Leaving it here rather than asking for review time on an unproven win. If anyone profiling on host-bound hardware — the L4 at 19–45% GPU utilisation in #1410 — wants to test it, the branch is `tuan3w:perf/assignment_device_unpack`, and the sync-count reproduction is a few lines with `torch.cuda.set_sync_debug_mode("warn")`. Happy to reopen if it shows up there.

One detail worth keeping regardless of this PR: `set_sync_debug_mode` attribution has to skip both `warnings.py` (≤3.13) and `_py_warnings.py` (3.14+). Missing the second name silently attributes every warning to the warnings machinery and makes such an assertion vacuous — which is exactly what happened to my first version of the test.

PR #1495, opened 2026-09-21
fix(export): keep fp16 CoreML exports on the Neural Engine and document the fallback boundary

lthphuw:
## Summary

RF-DETR ships two CoreML paths — a native `.mlpackage` and the ExecuTorch `backend="coreml"` delegate — and neither documented where Core ML stops using the Apple Neural Engine, nor had a test that would notice if that boundary moved. Issue #1024 lists both as "Medium" risk for exactly that reason ("ANE fallback" for native CoreML, "ANE partitioning" for the delegate).

This measures the boundary, documents it, pins it with tests, and fixes one ANE compile failure found along the way.

**The boundary**, measured with `MLComputePlan` on fp16 bundles under `CPU_AND_NE` (pretrained weights, Apple M3 Pro, macOS 27.0, coremltools 9.0):

| Model           | Ops on ANE | Ops on CPU | Share of estimated work on the ANE |
| --------------- | ---------- | ---------- | ---------------------------------- |
| `RFDETRNano`    | 592        | 7          | 99.9%                              |
| `RFDETRSmall`   | 650        | 8          | 99.9%                              |
| `RFDETRMedium`  | 708        | 8          | 99.9%                              |
| `RFDETRSegNano` | 729        | 8          | 99.9%                              |

The ops that leave the ANE are identical in every case: the two-stage `topk` plus the `expand_dims`/`tile`/`gather_along_axis` that consume its indices.

This is an fp16 property. At fp32 the plan reports *every* op as ANE-unsupported, and the same `RFDETRNano` bundle runs all 599 ops on the GPU under `ALL` and all 599 on the CPU under `CPU_AND_NE` — so fp32 never reaches the Neural Engine at all.

The ExecuTorch delegate takes the **whole** graph — no portable kernel call survives in the `.pte` — and that delegated model does reach the ANE (`com.apple.ane` unified-log activity, and 14.0 ms p50 where CPU is 73 ms and the fp32 GPU path is 34 ms). Its internal op split is *not* included above: `MLComputePlan` needs an `.mlpackage`, and the delegate blob inside a `.pte` is ExecuTorch's own container, which coremltools will not load.

**The compile failure.** Apple's ANE compiler rejects a whole fp16 program:

```
(ANECompiler) Invalid layer
Error Domain=com.apple.appleneuralengine.compiler Code=22
```

`CPU_AND_NE` then runs the entire model on CPU, and the default `ComputeUnit.ALL` cannot build an execution plan at all (`error code: -5`), so `predict` raises. Two conditions must hold together:

1. `enc_output_norm` still carries its **identity affine** (weight all ones, bias all zeros), and
2. its full-length output feeds the ANE-resident class head **and** crosses to the CPU-resident `topk`-indexed gather, at a token count that is a multiple of 32.

**Trained checkpoints are not affected**, which is why this never surfaced in normal use: the pretrained `RFDETRNano`, `RFDETRSmall`, `RFDETRSegSmall` and `RFDETRKeypointPreview` all compile for the ANE on `develop`. What this reaches is a model exported before training — the export test suite among them.

| Configuration (fp16 `RFDETRNano`, on `develop`)           | ANE compile |
| --------------------------------------------------------- | ----------- |
| pretrained                                                  | ok          |
| untrained (identity affine from default init), seeds 0/1/7  | fail        |
| pretrained, `enc_output_norm` forced to weight=1, bias=0    | fail        |
| untrained, `enc_output_norm` weight=1.001, bias=0           | ok          |
| untrained, `enc_output_norm` weight=1, bias=0.001           | ok          |

Narrowed by copying one pretrained parameter group at a time into an untrained model: only `enc_output_norm` flips the verdict. The MIL op is identical either way (`layer_norm` with gamma/beta constants present), so the compiler is reacting to the values. Non-finite constants and DINOv2 LayerScale were both ruled out.

The fix gathers the pre-norm rows and applies `enc_output_norm` to the selected tokens only. LayerNorm acts per token, so it is the same computation, and the norm's output no longer crosses the partition boundary.

## Changes

- `src/rfdetr/models/transformer.py`: the eval/export two-stage loop gathers `enc_output`'s output and normalizes the selected tokens. The batched training path (`_two_stage_group_selection`, `group_detr > 1`) is untouched; export always takes the loop.
- `docs/learn/export.md`: new "Neural Engine, Compute Units, and the Fallback Boundary" section — `compute_units` is a load-time choice and is not stored in the `.mlpackage` (Python and Swift snippets), fp32 never reaches the ANE while fp16 does, per-compute-unit latency, the boundary above, why `ALL` can be slower than `CPU_AND_NE` for segmentation (34.5 vs 23.5 ms on SegNano), the ~5 s first ANE load, and what fp16 costs in accuracy.
- `tests/export/test_coreml_ane.py` (new): fp16 load under `ComputeUnit.ALL` at 384 px (576 tokens, the rejected case) and 448 px (784, the control); an `MLComputePlan` test pinning the boundary; and one that catches a *silent* whole-model fallback, which neither the boundary test nor a load test can see.
- `tests/export/test_executorch_coreml.py` (new): the delegate's `.pte` must contain no portable kernel call, and must load and run. Nothing covered this path before — every other test mocks `executorch` away, and the `e2e_executorch` suite uses XNNPACK on ubuntu.
- `tests/models/test_transformer.py`: hardware-free guard asserting the gather never reads the full-length norm output.
- `.github/workflows/ci-integrations.yml`: new `executorch-coreml-parity` macOS job (wired into the guardian gate), and the new CoreML file added to `coreml-parity`.

Two things in the diff are separable from the two goals above, both found while writing the docs: the CoreML inference example passed `{"input": ...}` while the exported input is named `tensors` (a `KeyError` if copied verbatim), and `Transformer.forward` had no docstring, which the repo's convention check requires once the function is touched.

## Verification

Both new guards fail without the source change; on `develop` the e2e case behaves exactly as the token-count rule predicts (384 px fails, 448 px passes).

```
uv run --no-sync pytest src/ tests/ scripts/ -n 4 -m "not gpu and not coco17 and not integration and not xla and not tpu" \
  --ignore=tests/run_smoke_all_models.py --ignore=tests/legacy/test_checkpoint_compat.py
  # 5553 passed, 195 skipped, 3 failed
pytest tests/export/test_coreml_export.py tests/export/test_coreml_ane.py -m e2e_coreml -n 1          # 16 passed
pytest tests/export/test_executorch_coreml.py tests/export/test_executorch_export.py -m e2e_executorch -n 1  # 15 passed
pytest tests/export/test_litert_export.py -m e2e_litert -n 1                                          # 5 passed
pytest tests/export/ -m "e2e_onnxruntime or e2e_openvino" -n 2                                        # 5 passed, 3 skipped
pre-commit run --all-files                                                                            # passed
mkdocs build --strict                                                                                 # passed
```

The 3 failures are `test_openvino_export.py::TestExportOpenvinoMissingDependency*`; they fail identically on a clean `develop` checkout on this machine because OpenVINO 2026.4 is installed locally, and are unrelated to this change.

Not verified: the GPU suite and TensorRT parity (no CUDA device — this is an Apple M3 Pro), and CI has not run yet. The new ANE assertions need a host that exposes a Neural Engine; on a virtualized macOS runner the load test passes trivially and the compute-plan tests skip, so the non-vacuous guard there is the platform-independent graph test.

## Performance

Pretrained `RFDETRNano`, Apple M3 Pro, macOS 27.0, coremltools 9.0, batch 1. Accuracy is COCO val2017, all 5000 images, decoded through the repo's own `infer_transforms` / `post_process` / `CocoEvaluator`.

| Artifact            | mAP before | mAP after |
| ------------------- | ---------- | --------- |
| eager PyTorch CPU   | 48.024     | 48.024    |
| CoreML fp32 (`ALL`) | 48.023     | 48.023    |
| CoreML fp16 (`ALL`) | 45.054     | 45.059    |
| ExecuTorch xnnpack  | 48.022     | 48.022    |

Latency p50/p90 ms, median of 3 runs × 100 iterations, ≥10 warm-ups, 200 ms buffer between timed passes (the `docs/learn/benchmarks.md` methodology):

| Configuration               | before          | after           |
| --------------------------- | --------------- | --------------- |
| fp16 `CPU_AND_NE`           | 20.68 / 22.20   | 20.79 / 22.54   |
| fp16 `ALL`                  | 20.58 / 22.13   | 21.20 / 22.69   |
| fp16 `CPU_AND_GPU`          | 32.94 / 41.82   | 33.90 / 44.20   |
| fp16 `CPU_ONLY`             | 45.46 / 46.55   | 44.51 / 45.58   |
| fp32 `ALL`                  | 34.51 / 42.79   | 33.74 / 40.61   |
| fp32 `CPU_AND_GPU`          | 34.31 / 43.55   | 33.96 / 44.54   |
| fp32 `CPU_AND_NE`           | 73.92 / 76.26   | 72.90 / 74.28   |
| fp32 `CPU_ONLY`             | 73.72 / 75.65   | 73.62 / 75.48   |
| ExecuTorch xnnpack          | 180.90 / 192.26 | 180.34 / 184.68 |
| eager CPU (interleaved A/B) | 87.07 / 89.98   | 82.66 / 85.70   |

No cell regresses beyond noise; cold load is unchanged (~5.0 s for the ANE, ~0.5 s for GPU/CPU). The eager row was re-measured interleaved because a sequential pass read 78 ms before and 99 ms after — the machine heating up over the hour-long matrix, not a regression.

Two findings from the same measurements, documented but out of scope here: fp16 costs about 3 mAP on the native path (48.0 → 45.1), while the ExecuTorch delegate keeps 48.0 because it leaves the weights in fp32; and the ANE measures roughly twice as fast back-to-back (10.6 ms) as under the buffered methodology above, so any published absolute number has to state which was used.

## Related issues

Part of #1024.


.github/workflows/ci-integrations.yml
```diff
@@ -170,6 +173,67 @@ jobs:
...
+          assert _IS_EXECUTORCH_AVAILABLE, "executorch installed but not detected as available"
+          import coremltools  # noqa: F401
+
+          print("executorch runtime + coremltools import OK")
+        shell: python
+
+      - name: 🧪 Run ExecuTorch CoreML delegate tests
+        run: |
+          uv run --no-sync pytest tests/export/test_executorch_coreml.py \
+            -m e2e_executorch -n 1 \
```
Borda:
```suggestion
          uv run --no-sync pytest . -m e2e_executorch -n 1 \
```

Borda (approved):

